PR #26, opened 2026-03-27
fix: display names are changed

atakanbayrak:
## Description

Display names are changed with dot based definition. Storybook problems, tests, build are double-checked and no problem seen. 

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [x] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Checklist

## How to Test

Changes could be seen when local storybook checked.

### DIFF
--- a/src/lib/components/Grid/components/Col.stories.tsx
+++ b/src/lib/components/Grid/components/Col.stories.tsx
@@ -3,7 +3,7 @@ import type { Meta, StoryObj } from "@storybook/nextjs";
 import Grid from "@/components/Grid/Grid";
 
 const meta: Meta<typeof Grid.Col> = {
-  title: "Components/Grid/Col",
+  title: "Components/Grid/Grid.Col",
   component: Grid.Col,
   argTypes: {},
   args: {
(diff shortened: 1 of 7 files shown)


src/lib/components/Grid/components/Col.stories.tsx
```diff
@@ -3,7 +3,7 @@ import type { Meta, StoryObj } from "@storybook/nextjs";
 import Grid from "@/components/Grid/Grid";
 
 const meta: Meta<typeof Grid.Col> = {
-  title: "Components/Grid/Col",
+  title: "Components/Grid/Grid.Col",
```
aktasmehmet: This will cause the menu title of Col in Storybook. It is alreadt under Grid folder. So, I guess this is not necessary at all. You may want to check the same issue for other changes as well.

aktasmehmet (changes requested):

src/lib/components/Grid/components/Col.stories.tsx
atakanbayrak: That's true that this change changed the naming of menu title, but I thought that all naming should be matched even if the title. There is no problem to revert this change and will be reverted back to original.

aktasmehmet: You are right but in the menubar, it is already located under the parent component. So that would be a duplicate. Thx for the revert.

aktasmehmet (approved):

hatesalp (approved):

PR #27, opened 2026-03-27
fix: add self-error support for upload components within formfields

hatesalp:
## Description

- Eliminated duplicate error messages in UploadDragger and UploadList when used inside FormField
- File validation errors (size, type, upload/delete failures) were shown both in per-file labels (FileLabel) and in FormField helper text now only shown in FileLabel
- Added `suppressSelfErrorDisplay` option to `useRegisterFormField` for components that display their own errors internally
- FormField error styling (red border) is preserved via a separate `hasSelfError` boolean state
- Form-level validation (`isValid`) correctly tracks self-errors from upload components
- UploadInput behavior is unchanged  it still shows errors in FormField helper since it has no FileLabel
- Self-error tests have been added to UploadList/UploadDragger.

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [x] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [x] 🧹 Code Refactoring

## Screenshots / Preview

 **BEFORE**

{0F1F5DBB-8BBB-4350-8455-9051179789BA}

 **AFTER**

{146F86B9-4F6E-4D17-811C-EB6C3FB8CF76}

## Checklist

## How to Test
Example:
 ```
 <Form
        onSubmit={data => {
          if (data.isValid) {
            console.log("form geçerli!", data.values);
          } else {
            console.log("form geçersiz!");
          }
        }}
      >
        <FormField
          helperText="Please enter the application name"
          label="Application Name"
          name="appName1"
          validations={[Validations.RequiredUploadedFile]}
        >
          <UploadList
            accept={["*"]}
            maxSize={1000000}
            deleteRequest={{
              headers: [
                {
                  key: "mtf",
                  value: "ui",
                },
              ],
              method: "POST",
              url: "https://httpbin.org/post",
            }}
            uploadRequest={{
              headers: [
                {
                  key: "mtf",
                  value: "ui",
                },
              ],
              method: "POST",
              url: "https://httpbin.org/post",
            }}
          />
        </FormField>
        <FormField
          helperText="Please enter the application name"
          label="Application Name"
          name="appName"
          validations={[Validations.RequiredUploadedFile]}
        >
          <UploadDragger
            accept={["*"]}
            autoUpload
            deleteRequest={{
              headers: [
                {
                  key: "mtf",
                  value: "ui",
                },
              ],
              method: "POST",
              url: "https://httpbin.org/post",
            }}
            maxFile={10}
            maxSize={2}
            messages={{}}
            uploadRequest={{
              headers: [
                {
                  key: "mtf",
                  value: "ui",
                },
              ],
              method: "POST",
              url: "https://httpbin.org/post",
            }}
          />
        </FormField>
      </Form>
```

Closes [#EDKUI-1343]

### CODE REVIEW

**Comment by atakanbayrak** [comment]
I'll check this PR after mid break, it is not possible to look and understand the implementation on the first sight. sorry not enough energy 😢

File: src/lib/components/Upload/UploadDragger/UploadDragger.tsx
Comment (atakanbayrak) [inline src/lib/components/Upload/UploadDragger/UploadDragger.tsx:1]: I don't know how design team and our team decided to make changes on border collaring but it seems a little inappropriate to me when there is an error. There is a red border on the component but not in the little upload section which is under the upload area. My opinion if there is an error, red border should also cover this area. (I didn't find a correct place to write this comment so I wrote down here.)

File: src/lib/components/Form/context/FormContext.tsx [hunk 1]
```diff
@@ -194,6 +194,12 @@ export const FormProvider = (props: PropsWithChildren<FormProviderProps>) => {
       });
     });
 
+    Object.values(formStateRef.current.fields).forEach(field => {
+      if (field?.hasSelfError) {
+        isValid = false;
+      }
```
atakanbayrak:
```suggestion
isValid = !field?.hasSelfError;
```

src/lib/components/Form/context/FormContext.tsx
```diff
@@ -210,7 +216,12 @@ export const FormProvider = (props: PropsWithChildren<FormProviderProps>) => {
     formStateRef.current.values[nameToUpdate] = groupName
       ? { ...(formStateRef.current.values[nameToUpdate] as object), [name]: value }
       : value;
-    formStateRef.current.fields[nameToUpdate]?.errorSetter?.(undefined);
+    const field = formStateRef.current.fields[nameToUpdate];
+    if (field) {
+      field.errorSetter?.(undefined);
+      field.hasSelfError = false;
+      field.selfErrorSetter?.(false);
+    }
```
atakanbayrak, line 225:
```suggestion
    if (!field) return;
    
    field.errorSetter?.(undefined);
    field.hasSelfError = false;
    field.selfErrorSetter?.(false);
```

I think this is better for readability.

src/lib/components/Form/context/FormContext.tsx
```diff
@@ -219,10 +230,15 @@ export const FormProvider = (props: PropsWithChildren<FormProviderProps>) => {
...
-      formStateRef.current.fields[name]?.errorSetter?.(errorToShow);
+  const notifyFormForFieldSelfError = useCallback((name: string, errors: string[], suppressDisplay?: boolean) => {
+    const field = formStateRef.current.fields[name];
+    if (field) {
+      const hasError = errors.length > 0;
+      field.hasSelfError = hasError;
+      field.selfErrorSetter?.(hasError);
+      if (hasError && !suppressDisplay) {
+        field.errorSetter?.(errors.join("\n"));
+      }
```
atakanbayrak:
```suggestion
    if (!field) return;
    
    const hasError = errors.length > 0;
    field.hasSelfError = hasError;
    field.selfErrorSetter?.(hasError);
    if (hasError && !suppressDisplay) {
      field.errorSetter?.(errors.join("\n"));
    }
```

Same as upper comment.

src/lib/components/Form/types.ts
```diff
@@ -93,7 +94,7 @@ export type FormStateRefProps = {
 // Props that are provided to the fields via FieldContext
 export type FormContextType<T> = {
   notifyFormForFieldValueChange: (name: string, groupName: string | undefined, value?: InputValue) => void;
-  notifyFormForFieldSelfError: (name: string, errors: string[]) => void;
+  notifyFormForFieldSelfError: (name: string, errors: string[], suppressDisplay?: boolean) => void;
```
atakanbayrak: If this prop is same as "suppressSelfErrorDisplay", we should choose one of the naming instead of two.

atakanbayrak (changes requested):

atakanbayrak:
I think there is a really good work, appreciate it. 👏🏻

src/lib/components/Upload/UploadDragger/UploadDragger.tsx
aktasmehmet: This was by design and decided a long time ago. But I sort of agree with you. It would be better to open a task for this to handle in the future

aktasmehmet (commented):
I will leave a general comment here.

Overall, it is a really clever idea. You can improve it a little bit I guess. I've already sent the details but, in general:

- suppressSelfErrorDisplay, hasSelfError, setHasSelfError, notifyFormForFieldSelfError are not necessary
- You should not put error and selfError fields into FormField because it will make confusions. Only error would be enough I guess
- Main issue is in the validate function in FormContext. Actually it is probably an error from past. The async forEach loops there would cause some timing issues both for the validation check itself and the selfError stuff. I've sent you an updated version of it.

Thanks. When you refactor those, I'll check the changes in the test files.

src/lib/components/Upload/UploadDragger/UploadDragger.tsx
atakanbayrak: Done 👍🏻

hatesalp:
> I will leave a general comment here.
> 
> Overall, it is a really clever idea. You can improve it a little bit I guess. I've already sent the details but, in general:
> 
> * suppressSelfErrorDisplay, hasSelfError, setHasSelfError, notifyFormForFieldSelfError are not necessary
> * You should not put error and selfError fields into FormField because it will make confusions. Only error would be enough I guess
> * Main issue is in the validate function in FormContext. Actually it is probably an error from past. The async forEach loops there would cause some timing issues both for the validation check itself and the selfError stuff. I've sent you an updated version of it.
> 
> Thanks. When you refactor those, I'll check the changes in the test files.

Thanks for the information. I've made the necessary adjustments. You can check my last commit. 🙏

src/lib/components/Form/context/FormContext.tsx
hatesalp: It changed with my last commit. It was resolved in another way without needing these changes.

src/lib/components/Form/types.ts
hatesalp: It changed with my last commit. It was resolved in another way without needing these changes.

src/lib/components/Form/context/FormContext.tsx
hatesalp: It changed with my last commit. It was resolved in another way without needing these changes.

src/lib/components/Form/context/FormContext.tsx
hatesalp: It changed with my last commit. It was resolved in another way without needing these changes.

aktasmehmet (approved):
This approact is way cleaner and simpler. Great work 💪

atakanbayrak (approved):
That's really good work up there. Congrats.

PR #34, opened 2026-04-03
fix: support custom icon components in icon props for accordion avatar badge breadcrumb and businesscard

ZehranurC:
## Description

- `icon` prop now accepts both `string` (icon name) and `ReactElement` (e.g. `<FontAwesomeIcon />`) in Accordion, Avatar, Badge, Breadcrumb and BusinessCard components
- Updated rendering logic with `typeof icon === "string"` check
- Added tests for component icon usage
- Updated Storybook argTypes descriptions

## Type of Change

- [ ] 🆕 New Component
- [x] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Checklist

## How to Test

Closes [#EDKUI-1155]


src/lib/components/Accordion/Accordion.tsx
```diff
@@ -52,7 +52,7 @@ const AccordionComponent = (props: PropsWithRefAndChildren<AccordionProps, HTMLD
   return (
     <div className={classNames} style={style} ref={ref} data-testid="accordionItem">
       <button className={styles.header} onClick={toggleAccordion}>
-        {icon && <Icon size="lg" name={icon} />}
+        {icon && (typeof icon === "string" ? <Icon size="lg" name={icon} /> : icon)}
```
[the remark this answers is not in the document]
ZehranurC: The suggestion to wrap in <Icon> won't work as Icon is a Material Icons component, not a generic wrapper. Users passing a ReactElement are expected to handle sizing themselves, which is the standard pattern for custom icon props

src/lib/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -51,7 +51,11 @@ const Breadcrumb = (props: PropsWithRef<BreadcrumbProps, HTMLUListElement>) => {
       <ul className={classNames} ref={ref} style={style}>
         {itemsToRender.length > 0 &&
           (homeIcon ? (
-            <Icon name={homeIcon} className={styles["homepage-icon"]} />
+            typeof homeIcon === "string" ? (
+              <Icon name={homeIcon} className={styles["homepage-icon"]} />
+            ) : (
+              <span className={styles["homepage-icon"]}>{homeIcon}</span>
```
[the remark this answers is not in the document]
ZehranurC: Icon doesn't accept children, so the suggested change won't work. Users passing a ReactElement are expected to handle sizing themselves. this is the standard pattern for custom icon props.

src/lib/components/BusinessCard/BusinessCard.tsx
```diff
@@ -38,7 +38,7 @@ const BusinessCard = (props: PropsWithRef<BusinessCardProps, HTMLDivElement>) =>
   return (
     <div className={classNames} ref={ref} onClick={onClick} style={style}>
       {image && <Avatar image={image} size="xxl" />}
-      {icon && <Icon name={icon} className={styles.icon} />}
+      {icon && (typeof icon === "string" ? <Icon name={icon} className={styles.icon} /> : icon)}
```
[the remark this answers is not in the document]
ZehranurC: Added a test for the ReactElement path. The <span> wrapper is only applied to the ReactElement case — wrapping the string path too would add an unnecessary DOM element

src/lib/components/Accordion/Accordion.tsx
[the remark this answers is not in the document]
ZehranurC: Icon doesn't accept children so wrapping won't work. Sizing is the caller's responsibility when passing a ReactElement — this is the standard pattern for custom icon props.

src/lib/components/Badge/Badge.tsx
```diff
@@ -19,7 +19,19 @@ const Badge = (props: PropsWithRefAndChildren<BadgeProps, HTMLDivElement>) => {
     ref,
   } = usePropsWithThemeDefaults("Badge", props);
 
-  const contentToRender = icon ? <Icon name={icon} size="xxs" /> : content ? max > 0 && Number(content) > max ? max + "+" : content : null;
+  const contentToRender = icon ? (
+    typeof icon === "string" ? (
+      <Icon name={icon} size="xxs" />
+    ) : (
+      icon
+    )
```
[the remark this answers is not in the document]
ZehranurC: Icon doesn't accept children so wrapping won't work. Sizing is the caller's responsibility when passing a ReactElement — this is the standard pattern for custom icon props

src/lib/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -51,7 +51,11 @@ const Breadcrumb = (props: PropsWithRef<BreadcrumbProps, HTMLUListElement>) => {
...
+            )
```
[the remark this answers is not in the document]
ZehranurC: Valid point on semantics — however this is a pre-existing issue, not introduced by this PR. The home icon was already rendered directly under <ul> before this change. Happy to fix it in a separate PR.

src/lib/components/BusinessCard/BusinessCard.tsx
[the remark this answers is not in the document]
ZehranurC: Test has been added in the latest commit.

src/lib/components/Accordion/Accordion.tsx
aktasmehmet:
```suggestion
        {typeof icon === "string" ? <Icon size="lg" name={icon} /> : icon}
```
What about this? If the icon is not given, the type will be undefined. 

So 

if it is string -> renders <Icon>
else ->
    if it is undefined -> renders nothing
    if it is anything else -> renders anything else (external icons etc)

src/lib/components/Accordion/Accordion.test.tsx
```diff
@@ -40,6 +40,11 @@ describe("Accordion", () => {
     expect(screen.getByText("home")).toBeInTheDocument();
   });
 
+  it("should render icon component when icon prop is given as a component", () => {
+    render(<Accordion title="Accordion Title" icon={<span data-testid="custom-icon">★</span>} />);
+    expect(screen.getByTestId("custom-icon")).toBeInTheDocument();
+  });
+
```
aktasmehmet: I guess these 2 tests should be merged and test the string icon and element icon in the same test. The name should stay the same as "should display the icon given in the icon prop"

src/lib/components/Badge/Badge.tsx
```diff
@@ -19,7 +19,19 @@ const Badge = (props: PropsWithRefAndChildren<BadgeProps, HTMLDivElement>) => {
     ref,
   } = usePropsWithThemeDefaults("Badge", props);
 
-  const contentToRender = icon ? <Icon name={icon} size="xxs" /> : content ? max > 0 && Number(content) > max ? max + "+" : content : null;
+  const contentToRender = icon ? (
+    typeof icon === "string" ? (
+      <Icon name={icon} size="xxs" />
```
aktasmehmet: Maybe it would be better to create a const like:

```const maybeIcon = icon ? ....```

and use it inside contentToRender for better readability.

Besides, I also agree with Copilot's comment. In order to keep our icons and external icon consistent (in terms of size) we should give the size in css, not with size="xss"

**Review by aktasmehmet — COMMENTED** [review]
I've left my comments into Accordion component. Those comments also applies for other components as well.

File: src/lib/components/Accordion/Accordion.tsx (hunk 1 above)
Comment (ZehranurC) [reply src/lib/components/Accordion/Accordion.tsx]: You're right, undefined is already handled by the typeof check. I'll remove it.

File: src/lib/components/Accordion/Accordion.test.tsx (hunk 6 above)
Comment (ZehranurC) [reply src/lib/components/Accordion/Accordion.test.tsx]: Done, merged both into a single test.

File: src/lib/components/Badge/Badge.tsx (hunk 7 above)
Comment (ZehranurC) [reply src/lib/components/Badge/Badge.tsx]: Ok, extracted the icon logic into a maybeIcon const and moved the size to CSS.

File: src/lib/components/Avatar/Avatar.test.tsx [hunk 8]
```diff
@@ -14,6 +14,11 @@ describe("Avatar", () => {
     expect(render(<Avatar icon="info" />).getByText("info")).toBeInTheDocument();
   });
 
+  it("should render icon component when icon prop is given as a component", () => {
+    render(<Avatar icon={<span data-testid="custom-icon">★</span>} />);
```
atakanbayrak: Are there any solution without using data-testid, because there is an upcoming issue to delete all data-testid usage inside the project?

src/lib/components/Badge/Badge.test.tsx
```diff
@@ -84,6 +84,15 @@ describe("Badge", () => {
     expect(getByTestId(container, "badgeItem").textContent).not.toBe(testContent);
   });
 
+  it("should render icon component when icon prop is given as a component", () => {
+    const { container } = render(
+      <Badge icon={<span data-testid="custom-icon">★</span>}>
```
atakanbayrak: Same for the data-testid. Best to check for each component which was refactored under this PR.

atakanbayrak (changes requested):

src/lib/components/Accordion/Accordion.stories.tsx
```diff
@@ -6,6 +6,7 @@ const meta: Meta<typeof Accordion> = {
   component: Accordion,
   argTypes: {
     expanded: { table: { defaultValue: { summary: "false" } } },
+    icon: { description: "Icon name (string) or a custom icon component (ReactElement), e.g. `<FontAwesomeIcon />`" },
```
aktasmehmet:
```suggestion
    icon: { description: "Icon name (string) or a custom icon uses <span>, <i> or <svg>" },
```

We should also reflect this to other stories

src/lib/components/Badge/Badge.tsx
```diff
@@ -19,7 +19,9 @@ const Badge = (props: PropsWithRefAndChildren<BadgeProps, HTMLDivElement>) => {
     ref,
   } = usePropsWithThemeDefaults("Badge", props);
 
-  const contentToRender = icon ? <Icon name={icon} size="xxs" /> : content ? max > 0 && Number(content) > max ? max + "+" : content : null;
+  const maybeIcon = typeof icon === "string" ? <Icon name={icon} className={styles.icon} /> : icon;
+
+  const contentToRender = icon ? maybeIcon : content ? (max > 0 && Number(content) > max ? max + "+" : content) : null;
```
aktasmehmet: maybeIcon already is an icon or undefined. So, no need to use "icon" as the condition.

```suggestion
  const contentToRender = maybeIcon ?? content ? (max > 0 && Number(content) > max ? max + "+" : content) : null;
```

src/lib/components/BusinessCard/BusinessCard.test.tsx
```diff
@@ -22,9 +22,13 @@ describe("BusinessCard", () => {
...
   });
 
-  it("should render the main icon given in the icon prop", () => {
+  it("should display the icon given in the icon prop", () => {
     const { getByText } = render(<BusinessCard icon="folder" />);
     expect(getByText("folder")).toBeInTheDocument();
+
+    const IconComponent = () => <span>icon-svg</span>;
+    render(<BusinessCard icon={<IconComponent />} />);
+    expect(screen.getByText("icon-svg")).toBeInTheDocument();
```
aktasmehmet: screen is not imported before so I guess not necessary here. You can use rerender and just use getByText here to prevent unnecessary "screen" import

src/lib/components/Badge/Badge.tsx
```diff
@@ -19,7 +19,9 @@ const Badge = (props: PropsWithRefAndChildren<BadgeProps, HTMLDivElement>) => {
     ref,
   } = usePropsWithThemeDefaults("Badge", props);
 
-  const contentToRender = icon ? <Icon name={icon} size="xxs" /> : content ? max > 0 && Number(content) > max ? max + "+" : content : null;
+  const maybeIcon = typeof icon === "string" ? <Icon name={icon} className={styles.icon} /> : icon;
```
aktasmehmet: There are some fundemental issues in this PR and I want to state here:

In each component, the structure :

``const maybeIcon = typeof icon === "string" ? <Icon name={icon} className={styles.icon} /> : icon;``

is used. So this might be extracted as a component named "GlobalIconWrapper" under motif folder I guess.

Another and more important thing is that, when a custom icon is given, it cannot get the class name ("styles.icon" here) so the font size, color, etc. which are defined by the component itself is not applied to the custom button. 

So my suggestion is that in that GlobalIconWrapper component, custom component should be wrapped with a span with the given class name. 

One last thing. Most of the icons are rendered in span, i or svg. So all of these variants in terms of css shoul be handles. In addition to that, test should also cover that custom icons have the same class and visibility desired by the component.

src/lib/components/Avatar/Avatar.test.tsx
ZehranurC: In the previous commit, data-testid usage was removed and replaced with an alternative approach

src/lib/components/Badge/Badge.test.tsx
ZehranurC: In the previous commit, data-testid usage was removed and replaced with an alternative approach.

src/lib/components/Accordion/Accordion.test.tsx
```diff
@@ -38,6 +38,10 @@ describe("Accordion", () => {
   it("should display the icon given in the icon prop", () => {
     render(<Accordion title="Accordion Title" icon="home" />);
     expect(screen.getByText("home")).toBeInTheDocument();
+
+    render(<Accordion title="Accordion Title" icon={<span>★</span>} />);
+    expect(screen.getByText("★")).toBeInTheDocument();
+    expect(screen.getByText("★").parentElement).toHaveClass("icon");
   });
```
atakanbayrak: We should specifice to query for looking inside component not screen maybe ?

atakanbayrak (changes requested):

src/lib/components/Accordion/Accordion.test.tsx
aktasmehmet: I was thinking the same but since the test was already using screen before, It's up to @ZehranurC I guess

src/lib/components/Motif/Icon/components/GlobalIconWrapper.tsx
```diff
@@ -0,0 +1,19 @@
+import { ComponentProps, ReactElement } from "react";
+import Icon from "../../../Icon/Icon";
+
+interface GlobalIconWrapperProps {
+  icon: string | ReactElement;
+  className?: string;
+  iconProps?: Omit<ComponentProps<typeof Icon>, "name" | "className">;
```
aktasmehmet: iconProps is not necessary I guess.

src/lib/components/Motif/Icon/components/GlobalIconWrapper.tsx
```diff
@@ -0,0 +1,19 @@
+import { ComponentProps, ReactElement } from "react";
+import Icon from "../../../Icon/Icon";
+
+interface GlobalIconWrapperProps {
```
aktasmehmet: we dont use interface. only "type"

src/lib/components/Motif/Icon/components/GlobalIconWrapper.tsx
```diff
@@ -0,0 +1,19 @@
...
+  icon: string | ReactElement;
+  className?: string;
+  iconProps?: Omit<ComponentProps<typeof Icon>, "name" | "className">;
+}
+
+const GlobalIconWrapper = ({ icon, className, iconProps }: GlobalIconWrapperProps) =>
+  typeof icon === "string" ? (
+    <Icon name={icon} className={className} {...iconProps} />
+  ) : (
+    <span className={className} style={{ display: "inline-flex", alignItems: "center" }}>
```
aktasmehmet: inline styling is not necessary here. you may create a seperate styles.module.scss file and put those inline styling there. You may use "externalIcon" name for the style.

Besides, as I mention in other comments, you can also move 

```
    > svg {
      width: 1em;
      height: 1em;
    }
```

here in this file as well.

src/lib/components/Motif/Icon/components/GlobalIconWrapper.tsx
```diff
@@ -0,0 +1,19 @@
+import { ComponentProps, ReactElement } from "react";
```
aktasmehmet: This component is related but not included into /Motif/Icon. So, moving this into its seperate forlder under /Motif/GlobalIconWrapper whould be better I guess. It will has its own style file, etc.

src/lib/components/BusinessCard/BusinessCard.module.scss
```diff
@@ -12,6 +12,10 @@ $variants: (("neutral", "light"), ("primary", "primary"), ("secondary", "seconda
 
   .icon {
     font-size: var(--base-sizing-40x);
+    > svg {
+      width: 1em;
+      height: 1em;
+    }
```
aktasmehmet: this section is used in every component which uses GlobalIconWrapper. So it would be better to move this into GlobalIconWrapper's own style file.

I dont comment on other similar sections. You will fix other components as well.

src/lib/components/Avatar/Avatar.test.tsx
```diff
@@ -10,8 +10,13 @@ describe("Avatar", () => {
     expect(render(<Avatar letters="test" />).getByText("TE")).toBeInTheDocument();
   });
 
-  it("should render icon when icon is set", () => {
-    expect(render(<Avatar icon="info" />).getByText("info")).toBeInTheDocument();
+  it("should display the icon given in the icon prop", () => {
```
aktasmehmet: All of those tests are the same. So, I guess that would be a better approach to move this tests under GlobalIconWrapper as its test. Somehow export that and use here, in these components (like DatePicker DateTimePicker etc.)

The test should include:
- name
- component check (svg or i or span)

src/lib/components/Motif/Icon/components/GlobalIconWrapper.tsx
```diff
@@ -0,0 +1,19 @@
...
+const GlobalIconWrapper = ({ icon, className, iconProps }: GlobalIconWrapperProps) =>
+  typeof icon === "string" ? (
+    <Icon name={icon} className={className} {...iconProps} />
+  ) : (
+    <span className={className} style={{ display: "inline-flex", alignItems: "center" }}>
+      {icon}
+    </span>
+  );
+
+export default GlobalIconWrapper;
```
aktasmehmet: I write the most important thing here:

This component is very dangerous in terms of the outcome. So, we should be certain that it renders the same results all the time. We cannot do that with jest, etc. So, we should pass this to chromatic.


In order to do that:

- You should create a stories file for GlobalIconWrapper
- This file should be private. It should not be shown in the storybook itself (But it should be uploaded to Chromatic. Please check this possibility before starting to work)
- There should be seperate stories for each component that uses GlobalIconWrapper. For example, for this PR, there should be 4 stories
- Each story should render the icons in the component (Avatar, Badge, etc) in 5 different ways:

```
<Avatar size={size} icon="folder" />
<Avatar size={size} icon={<i className="bi bi-amazon" />} />
<Avatar size={size} icon={<span className="material-symbols-outlined">android</span>} />
<Avatar
        size={size}
        icon={
          <svg
            xmlns="http://www.w3.org/2000/svg"
            width="16"
            height="16"
            fill="currentColor"
            className="bi bi-brightness-high"
            viewBox="0 0 16 16"
          >
            <path d="M8 11a3 3 0 1 1 0-6 3 3 0 0 1 0 6m0 1a4 4 0 1 0 0-8 4 4 0 0 0 0 8M8 0a.5.5 0 0 1 .5.5v2a.5.5 0 0 1-1 0v-2A.5.5 0 0 1 8 0m0 13a.5.5 0 0 1 .5.5v2a.5.5 0 0 1-1 0v-2A.5.5 0 0 1 8 13m8-5a.5.5 0 0 1-.5.5h-2a.5.5 0 0 1 0-1h2a.5.5 0 0 1 .5.5M3 8a.5.5 0 0 1-.5.5h-2a.5.5 0 0 1 0-1h2A.5.5 0 0 1 3 8m10.657-5.657a.5.5 0 0 1 0 .707l-1.414 1.415a.5.5 0 1 1-.707-.708l1.414-1.414a.5.5 0 0 1 .707 0m-9.193 9.193a.5.5 0 0 1 0 .707L3.05 13.657a.5.5 0 0 1-.707-.707l1.414-1.414a.5.5 0 0 1 .707 0m9.193 2.121a.5.5 0 0 1-.707 0l-1.414-1.414a.5.5 0 0 1 .707-.707l1.414 1.414a.5.5 0 0 1 0 .707M4.464 4.465a.5.5 0 0 1-.707 0L2.343 3.05a.5.5 0 1 1 .707-.707l1.414 1.414a.5.5 0 0 1 0 .708" />
          </svg>
        }
      />
<Avatar size={size} icon={<Camera />} />
```

- Those stories should include the most absurd rendering of the icons. I mean, different colors (to check if color change effects), etc.

Then, when any change happens in GlobalItemWrapper or the component that uses that, Chromatic will check the differences. If there is an issue, that it will report. So, we will not miss an error there.

I also attach the render of the code above:

Image

aktasmehmet (changes requested):

src/lib/components/Badge/Badge.tsx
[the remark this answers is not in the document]
ZehranurC: This section has been refactored

src/lib/components/Badge/Badge.module.scss
```diff
@@ -22,6 +22,10 @@ $positions: ("top-left", "top-right", "bottom-left", "bottom-right");
   vertical-align: middle;
 }
 
+.icon {
+  font-size: var(--base-sizing-16x);
```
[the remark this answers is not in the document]
ZehranurC: This section has been refactored

src/lib/components/Accordion/Accordion.test.tsx
ZehranurC: This section has been refactored

src/lib/components/Motif/Icon/components/GlobalIconWrapper.tsx
ZehranurC: This section has been refactored

src/lib/components/Motif/Icon/components/GlobalIconWrapper.tsx
ZehranurC: This section has been refactored

src/lib/components/Motif/Icon/components/GlobalIconWrapper.tsx
ZehranurC: This section has been refactored

src/lib/components/Motif/Icon/components/GlobalIconWrapper.tsx
ZehranurC: This section has been refactored

src/lib/components/BusinessCard/BusinessCard.module.scss
ZehranurC: This section has been refactored

src/lib/components/Avatar/Avatar.test.tsx
ZehranurC: A test file was created for GlobalIconWrapper

src/lib/components/Motif/Icon/components/GlobalIconWrapper.tsx
ZehranurC: Added Chromatic stories for all 5 affected components covering all supported icon types

src/lib/components/BusinessCard/BusinessCard.module.scss
```diff
@@ -12,6 +12,10 @@ $variants: (("neutral", "light"), ("primary", "primary"), ("secondary", "seconda
...
   }
```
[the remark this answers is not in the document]
ZehranurC: This section has been refactored

src/lib/components/Breadcrumb/Breadcrumb.module.scss
```diff
@@ -41,5 +41,10 @@
   .homepage-icon {
     color: var(--theme-color-text-primary-default);
     margin-right: var(--base-sizing-8x);
+    font-size: var(--base-sizing-22x);
+    > svg {
+      width: 1em;
+      height: 1em;
+    }
   }
```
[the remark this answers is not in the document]
ZehranurC: This section has been refactored

src/lib/components/Motif/Icon/components/GlobalIconWrapper.tsx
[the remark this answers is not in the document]
ZehranurC: This section has been refactored

src/lib/components/Avatar/Avatar.module.scss
```diff
@@ -45,6 +45,10 @@ $variants: ("primary", "secondary", "info", "success", "warning", "danger");
     }
     &.icon span {
       font-size: var(--base-sizing-#{$iconSize});
+      > svg {
+        width: 1em;
+        height: 1em;
+      }
     }
```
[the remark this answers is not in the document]
ZehranurC: This section has been refactored

src/lib/components/Badge/Badge.module.scss
```diff
@@ -22,6 +22,14 @@ $positions: ("top-left", "top-right", "bottom-left", "bottom-right");
...
+  > svg {
+    width: 1em;
+    height: 1em;
+  }
+}
```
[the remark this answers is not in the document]
ZehranurC: This section has been refactored

src/lib/components/Accordion/Accordion.module.scss
```diff
@@ -53,6 +53,14 @@
     }
   }
 
+  .icon {
+    font-size: var(--base-sizing-24x);
+    > svg {
+      width: 1em;
+      height: 1em;
+    }
+  }
```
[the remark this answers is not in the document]
ZehranurC: This section has been refactored

src/lib/components/BusinessCard/BusinessCard.tsx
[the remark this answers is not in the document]
ZehranurC: This section has been refactored

src/lib/components/Accordion/Accordion.test.tsx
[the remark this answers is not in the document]
ZehranurC: This section has been refactored

src/lib/components/Motif/GlobalIconWrapper/GlobalIconWrapper.tsx
```diff
@@ -0,0 +1,17 @@
...
+type GlobalIconWrapperProps = {
+  icon: string | ReactElement;
+  className?: string;
+};
+
+const GlobalIconWrapper = ({ icon, className }: GlobalIconWrapperProps) =>
+  typeof icon === "string" ? (
+    <Icon name={icon} className={className} />
+  ) : (
+    <span className={`${styles.externalIcon}${className ? ` ${className}` : ""}`}>{icon}</span>
```
aktasmehmet: You may use sanitize method here for the class names

src/lib/components/Motif/GlobalIconWrapper/GlobalIconWrapper.test.tsx
```diff
@@ -0,0 +1,27 @@
+import "@testing-library/jest-dom";
+import { render, RenderResult } from "@testing-library/react";
+import GlobalIconWrapper from "./GlobalIconWrapper";
+import { ReactElement } from "react";
+
+export type IconRender = (icon: string | ReactElement) => RenderResult;
+
+export const runIconPropTest = (renderIcon: IconRender, className?: string, iconName = "home") => {
```
aktasmehmet: 3rd parameter iconName here is never set when this method is called. So it would be better to but it as a const inside the function, instead of using as a parameter.

aktasmehmet: maybe ```render: ComponentWithIconRender``` would be a better naming.

src/lib/components/Motif/GlobalIconWrapper/GlobalIconWrapper.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+  it("should render string icon with correct name and element type", () => {
+    const { getByText } = renderIcon(iconName);
+    const iconElement = getByText(iconName);
+    expect(iconElement.tagName.toLowerCase()).toBe("span");
+    className && expect(iconElement).toHaveClass(className);
+  });
+
+  it("should render ReactElement icon with correct element type and wrapper class", () => {
```
aktasmehmet: I guess keeping ```it()...``` blocks inside each components own test would be a better idea.

For example you can convert 

```  it("should render ReactElement icon with correct element type and wrapper class", () => {```

here into

```const testIconElement = () => {```

and call ```testIconElement()``` below

File: src/lib/components/BusinessCard/BusinessCard.test.tsx [hunk 31]
```diff
@@ -22,10 +23,7 @@ describe("BusinessCard", () => {
     expect(getByText("description")).toBeInTheDocument();
   });
 
-  it("should render the main icon given in the icon prop", () => {
-    const { getByText } = render(<BusinessCard icon="folder" />);
-    expect(getByText("folder")).toBeInTheDocument();
-  });
+  runIconPropTest(icon => render(<BusinessCard icon={icon} />), "icon");
```
aktasmehmet: You may keep the test ```should render the main icon given in the icon prop``` here and call ```runIconPropTest()``` in it. 

I dont write this in other test files. You can update them.

aktasmehmet (changes requested):
I've just sanitized the new stories a little bit. It looks more refined now. 

The thing is that, some of them shows that the usage is incorrrect. Actually it helped us to detect some errors. We have to fix them.

Badge and BusinessCard has some issues when rendering Material Icons.

Image
Image

aktasmehmet (approved):

hatesalp (approved):

atakanbayrak (approved):

PR #37, opened 2026-04-07
feat: created timeline component

hatesalp:
## Description

Added a new `<Timeline />` component 

- **Orientation support**: Renders in both `vertical` and `horizontal` layouts
- **Marker types**: Supports three marker styles -->  `dot`, `number`, and `icon`
- **Content positioning**: Items can be placed `before`, `after`, or `alternate` relative to the timeline axis
- **Text alignment**: Configurable text alignment with `start`, `center`, and `end` options
- **Variants**: Six color variants available  --> `primary`, `secondary`, `success`, `warning`, `danger`, and `light`
- **Per-item customization**: Each timeline item can override its own `variant`, `icon`, and `appearance` (`filled` / `outlined`)
- **Disabled state**: Individual items can be marked as disabled
- **Flexible content**: Supports `ReactNode` titles and string content per item
- **Storybook coverage**: Includes stories for both `Timeline` and `TimelineItem` covering all variants and configurations

## Type of Change

- [x] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Screenshots / Preview

{78AAB1F2-CBD6-4A4A-B107-F9B00E6C0579}

{7659F0BD-9060-4BF8-95AD-653E3DC0C713}

{F17A3BE9-ECAF-4434-A5E3-13B1EDBB123D}

## Checklist

## How to Test

Closes [#EDKUI-756]: Internal purposes only

### DIFF
--- a/src/lib/components/Timeline/Timeline.module.scss
+++ b/src/lib/components/Timeline/Timeline.module.scss
@@ -57,8 +57,6 @@ $variants: (
 
 .Root {
   display: flex;
-  height: min-content;
-  width: fit-content;
 }
 
 .item {
--- a/src/lib/components/Timeline/Timeline.module.scss
+++ b/src/lib/components/Timeline/Timeline.module.scss
@@ -58,6 +58,7 @@ $variants: (
 .Root {
   display: flex;
   height: min-content;
+  width: fit-content;
 }
 
 .item {
--- a/src/lib/components/Timeline/Timeline.module.scss
+++ b/src/lib/components/Timeline/Timeline.module.scss
@@ -1,0 +1,311 @@
+@use "@styles" as *;
+
+// variant, fillTextColor
+$variants: (
+  ("primary", "base-global-light"),
+  ("secondary", "base-global-light"),
+  ("success", "base-global-light"),
+  ("warning", "base-text-dark"),
+  ("danger", "base-global-light"),
+  ("light", "light-default")
+);
+
+@mixin connection-line($orientation, $half: false) {
+  content: "";
+  display: block;
+  flex: 1;
+  background-color: var(--theme-color-border-light-default);
+
+  @if $orientation == "vertical" {
+    min-height: if($half, var(--base-sizing-12x), var(--base-sizing-24x));
+    width: var(--base-sizing-1x);
+  } @else if $orientation == "horizontal" {
+    min-width: if($half, var(--base-sizing-12x), var(--base-sizing-24x));
+    height: var(--base-sizing-1x);
+  }
+}
+
+@mixin marker-variant($variant, $fill-text-token, $self: false) {
+  $base: if($self, "&", "");
+  $outlined-disabled: if($self, "&.disabled.outlined", ".item.disabled.outlined");
+  $filled-disabled: if($self, "&.disabled.filled", ".item.disabled.filled");
+
+  #{$base}.outlined .marker span {
+    border-color: var(--theme-color-border-#{$variant}-default);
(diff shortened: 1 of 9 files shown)


src/lib/components/Timeline/Timeline.module.scss
```diff
@@ -0,0 +1,312 @@
+@use "@styles" as *;
+
+// variant, fillTextColor
+$variants: (
+  ("primary", "base-global-light"),
+  ("secondary", "base-global-light"),
+  ("success", "base-global-light"),
+  ("warning", "base-text-dark"),
+  ("danger", "base-global-light"),
+  ("light", "light-default")
```
atakanbayrak, line 10: We could define these color values under theme based values.

src/lib/components/Timeline/Timeline.module.scss
```diff
@@ -0,0 +1,312 @@
...
+}
+
+.Root {
+  display: flex;
+  height: min-content;
+  width: fit-content;
+}
+
+.item {
+  display: flex;
```
atakanbayrak, line 63: Root has display flex so .item and .section has it automatically I guess.

src/lib/components/Timeline/Timeline.module.scss
```diff
@@ -0,0 +1,312 @@
...
+
+.vertical {
+  &,
+  .marker {
+    flex-direction: column;
+  }
+
+  .item {
+    padding-left: var(--base-sizing-8x);
+    padding-right: var(--base-sizing-8x);
```
atakanbayrak, line 122:
```suggestion
padding: 0 var(--base-sizing-8x);
```

src/lib/components/Timeline/TimelineItem.tsx
```diff
@@ -0,0 +1,38 @@
...
+      {(title || content) && (
+        <div className={styles.section}>
+          {title && <span className={styles.title}>{title}</span>}
+          {content && <span className={styles.content}>{content}</span>}
+        </div>
+      )}
+    </div>
+  );
+});
+
```
atakanbayrak, line 37: `displayName` could be defined as "TimelineItem" for storybook naming conversion.

atakanbayrak (changes requested):

src/lib/components/Timeline/Timeline.module.scss
aktasmehmet: They alredy are. These are partial tokens that are used in full token names.  Please refer to ->     background-color: var(--theme-color-surface-#{$variant}-default);

src/lib/components/Timeline/Timeline.module.scss
aktasmehmet: I don't think this is the case. They should be given seperately.

src/lib/components/Timeline/TimelineItem.tsx
aktasmehmet: this is not an external component. it is just a json object. So it is not needed here.

src/lib/components/Timeline/Timeline.module.scss
aktasmehmet: since the scss file is too complicated and there are other paddings added to the item and its content, somewhere in the code, this is given seperately on purpose.

aktasmehmet (approved):

atakanbayrak (approved):

PR #38, opened 2026-04-08
refactor: general refactoring clean up

atakanbayrak:
## Description

General refactoring and cleanup is completed. Changes should be discussed in advance.

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [x] 🧹 Code Refactoring

## Checklist

## How to Test

Closes [#EDKUI-1232]


src/utils/utils.ts
```diff
@@ -53,8 +53,6 @@ export const isNotAvailable = (value: unknown) =>
  */
 export const camelToKebab = (str: string) => str.replace(/[A-Z]/g, letter => `-${letter.toLowerCase()}`);
 
-export const compareByDiffWithFlooringValues = (n1: number, n2: number, diffValue: number) => Math.floor(n1) - Math.floor(n2) > diffValue;
```
ZehranurC, line 56: compareByDiffWithFlooringValues was removed from utils.ts but its documentation still exists in utilityFunctions.mdx. You could also remove the corresponding section from the docs to keep them in sync.

atakanbayrak: Nice catch 🫡 . New commit is sent.

.storybook/MotifDoc/MotifDocContainer.tsx
```diff
@@ -1,7 +1,7 @@
 import { DocsContainer } from "@storybook/addon-docs/blocks";
 import "../../src/lib/styles/themes/default-theme.css";
 import { PropsWithChildren } from "react";
-import "./styles.scss";
+import "./MotifDocStyles.scss";
```
aktasmehmet: This is the folder structure that we use in every component of us. So, I guess keeping as "styles" would be better.

src/docs/components/CSSDocumentation/components/CSSClassNames.module.scss
```diff
@@ -0,0 +1,14 @@
...
+
+  .gridItem {
+    padding: 6px 10px;
+    background-color: var(--theme-color-surface-base-base2);
+    border-radius: 6px;
+    font-size: 13px;
+    border: 1px solid var(--theme-color-border-light-default);
+    color: var(--theme-color-text-base-title, #333333);
+  }
+}
```
aktasmehmet, line 14: Can you put a screenshot of the change here, in the description? Thx.

.storybook/main.ts
```diff
@@ -4,22 +4,21 @@ import { applyCustomCSSModuleNaming } from "../src/lib/styles/scripts/build";
...
-  core: {
-    disableTelemetry: true,
-  },
+  core: { disableTelemetry: true },
 
   addons:
     process.env.NODE_ENV === "production"
       ? ["@storybook/addon-links", "@storybook/addon-docs"]
-      : ["@storybook/addon-links", "@chromatic-com/storybook", "@storybook/addon-docs"],
+      : ["@storybook/addon-links", "@storybook/addon-docs", "@chromatic-com/storybook"],
```
aktasmehmet, line 15: Are you sure you deeply tested here? Sometimes some addons are meant to stay at a certain position (I am not sure about here). Just, a simple refactor may cause a significant issue. fyi.

aktasmehmet (changes requested):
Since some files are deleted, I have to state some of my comments here:

1 - I am not a fan of moving the data and mock from docs section into the component itself. Sometimes that data might be used somewhere else and it is mostly document related. This approach may cause misunderstanding that there is something about the data in the component itself.

So, that would be better to revert the folder changes for "src/docs/data/table.ts" and 
"src/docs/data/mock.ts"

2 - Motif Icon README file is necessary. It is used as a guide to change the icons. You just need to refactor the text and content in case of finding an internal text like a key or something. You should bring it back 😉

aktasmehmet:
This is not a "feat" I guess. Can you change the title prefix to "refactor" or something related?

src/docs/components/CSSDocumentation/components/CSSClassNames.module.scss
atakanbayrak: Screenshot 2026-04-13 at 09 40 42

Naming is the only change in here.

.storybook/main.ts
atakanbayrak: I double-checked this change, and there was no problem. Checked again rn, still no problem.

aktasmehmet (dismissed):

aktasmehmet (approved):

ZehranurC (approved):

PR #39, opened 2026-04-08
chore(build): strip data-testid attributes from production library bundle

hatesalp:
## Description

- data-testid attributes, which provide convenience for test components, are unnecessarily added to the DOM even after the build.

- Integrates `rollup-plugin-jsx-remove-attributes` into the rollup pipeline. Source files are not modified  tests remain unaffected.

- The library is published from `dist/` (produced by `rollup -c`), not from `next build`, so the cleanup must happen in the rollup pipeline. A Next.js `reactRemoveProperties` compiler option would only affect our internal demo site, not the npm package consumers.

- The `removeJsxAttributes()` attribute defaults to `data-testid` for readability, but I still wanted to mention it. Other attributes can be added in the future.

 **`rollup.config.mjs`** : integrates `rollup-plugin-jsx-remove-attributes` into the library pipeline. `NODE_ENV` is defaulted to `"production"`  at the top of the config (using `process.env.NODE_ENV || "production"`) so the plugin runs on every platform without requiring `cross-env` or changes to the npm script.

## Results

 Metric | Before | After |
|---|---|---|
| `data-testid` matches in `dist/index.es.js` | 54 | **0** |
| `data-testid` matches in `dist/index.cjs` | 54 | **0** |
| `dist/index.es.js` size | 620,453 B | 618,855 B |
| `dist/index.cjs` size | 629,844 B | 628,246 B |
| **Total saved** | — | **~3.1 KB** |

--------------------------------------------------

When Storybook uses the @storybook/nextjs framework, it inherits Next.js's SWC compiler. The Babel pipeline is disabled while SWC is active; therefore, no Babel plugins were used.

Storybook uses Next.js's built-in `compiler.reactRemoveProperties` property, which is naturally passed to Next.js SWC.

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Checklist

## How to Test
You can comment out `removeJsxAttributes()` and run the following commands sequentially to compare the before-and-after results.
```
npm run build:lib
dir dist\index.es.js dist\index.cjs
```
```
npm run build:storybook
npm run storybook
```

 Closes [#EDKUI-1233]: Internal purposes only


package.json
```diff
@@ -75,6 +75,7 @@
     "rollup": "4.59.0",
     "rollup-plugin-delete": "^3.0.2",
     "rollup-plugin-dts": "6.3.0",
+    "rollup-plugin-jsx-remove-attributes": "^3.1.2",
```
atakanbayrak: `^` symbol should be deleted.

rollup.config.mjs
```diff
@@ -9,9 +9,13 @@ import copy from "@rollup-extras/plugin-copy";
 import url from "postcss-url";
 import del from "rollup-plugin-delete";
 import alias from "@rollup/plugin-alias";
+import removeJsxAttributes from "rollup-plugin-jsx-remove-attributes";
 import path from "path";
+import process from "node:process";
 import { fileURLToPath } from "url";
 
+process.env.NODE_ENV = process.env.NODE_ENV || "production";
```
atakanbayrak, line 17: I saw that process is also used under /storybook/main.ts but there were no import line for it. It has been used statically, is it also possible to use this on that way ?

atakanbayrak (changes requested):

rollup.config.mjs
hatesalp: In `.mjs` files (strict ESM), `process` is not available as a global, so it should be explicitly imported using `import process from "node:process"`.
However, in `.ts` files (such as Next.js or Storybook configs), `process` is automatically provided by the Node.js runtime since they run in a CommonJS-compatible environment.

atakanbayrak: Nice information, thanks. Resolved.

next.config.ts
```diff
@@ -8,6 +8,10 @@ const nextConfig: NextConfig = {
     silenceDeprecations: ["if-function"],
   },
   turbopack: {},
+  compiler: {
+    reactRemoveProperties:
+      process.env.STORYBOOK === "true" && process.env.NODE_ENV === "production" ? { properties: ["^data-testid$"] } : false,
+  },
```
atakanbayrak, line 14: I guess that data-testid extracting only applies on storybook and production environment right ? I stated that only this change differs from previous commit ?

hatesalp: Initially, the code was only valid for the production environment, but after consultation with @aktasmehmet. I added it to the storybook production environment as well. Since we closed the test areas for the local Storybook, I developed the code to compile only in the production environment.

aktasmehmet (approved):

ZehranurC (approved):

atakanbayrak (approved):

PR #40, opened 2026-04-09
fix: workflow files have been refactored

atakanbayrak:
## Description

Workflow files have been refactored for avoiding fork attacks and injected workflow files implementation.

## Type of Change

- [ ] 🆕 New Component
- [x] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Checklist

## How to Test

Fork PR is tested in demo environment.


.github/workflows/chromatic_ui.yml
```diff
@@ -1,4 +1,4 @@
-name: UI
```
aktasmehmet, line 1: This naming was intentional. The running job names look horrible, so it is grouped under UI. Otherwise, it will be something like "Chromatic UI / Chromatic"

.github/workflows/workflow_protection.yml
```diff
@@ -0,0 +1,64 @@
+name: "Workflow Protection"
+
+on:
+  pull_request_target:
+    types: [opened, synchronize]
```
aktasmehmet: are those enough or should we put ``[opened, edited, reopened, synchronize]``?

.github/workflows/workflow-protection.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Check for unauthorized .github/ changes
+        uses: actions/github-script@v7
+        with:
+          script: |
+            // 1. Get all files changed in this PR via API (no git fetch needed)
+            const { data: files } = await github.rest.pulls.listFiles({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: context.payload.pull_request.number,
+              per_page: 100,
```
aktasmehmet, line 32: in theory, there is a possibility to miss the thing in very large PRs. Yu may need to paginate here something like:

```
script: |
  let page = 1;
  let allFiles = [];
  while (true) {
    const { data: files } = await github.rest.pulls.listFiles({
      owner: context.repo.owner,
      repo: context.repo.repo,
      pull_number: context.payload.pull_request.number,
      per_page: 100,
      page,
    });
    allFiles = allFiles.concat(files);
    if (files.length < 100) break;
    page++;
  }

  const sensitiveFiles = allFiles.filter(f =>
    f.filename.startsWith('.github/workflows/') ||
    f.filename.startsWith('.github/actions/')
  );
  // ...
```

.github/workflows/pr_validator.yml
aktasmehmet, line 25: The only problematic (sort of) section is here I guess. Pull request head will load the fork code here. So the fork user may bypass the core team output in the check-core-team/action.yml file just like:

```
runs:
  using: composite
  steps:
    - run: echo "is-internal=true" >> $GITHUB_OUTPUT
      shell: bash
```

So if I think correctly, this will allow the attacker to behave like a core team member along with other actions. Can you double check this with a similar try?

aktasmehmet (changes requested):
Along with other requests.... **kebab-case naming is preferred here. So please revert some of the naming changes and rename if there are other incorrect namings 🙏**

.github/workflows/workflow_protection.yml
atakanbayrak: edited only works when PR body is changed and it has no business with changed code itself so only reopened added. No edit needed.

.github/workflows/pr_validator.yml
atakanbayrak: When attacker changes the any of the workflow files it will appear on the workflow, and approve needed for PR validator, this change will remain silent unless we approve.

aktasmehmet: I am still not sure.

workflow-protection.yml is a status check, it can block a merge, but it runs alongside pr-validator.yml, not before it. Both workflows trigger simultaneously when a PR is opened. 

So:
- PR opens with malicious check-core-team
- Both workflows start at the same time
- pr-validator.yml loads and runs the attacker's check-core-team — bypass already happened
- workflow-protection.yml eventually fails and blocks the merge

So, this might solve the issue I guess:

# Before
```
- uses: actions/checkout@v4
  with:
    repository: ${{ github.event.pull_request.head.repo.full_name }}
    ref: ${{ github.event.pull_request.head.ref }}
    persist-credentials: 'false'
    fetch-depth: 0
```

# After
```
- uses: actions/checkout@v4
  with:
    persist-credentials: 'false'
```

atakanbayrak: I understand that it runs alongside but it is not possible to start before we give approve to workflows to run. I can still make this change that you refer but I want to be sure that we are on the same page.

aktasmehmet: Yep got it. Then lets keep it as it is. This is not a possibly fatal thing at all. I close this one 👍

aktasmehmet (approved):

PR #47, opened 2026-04-15
feat: add multi-language (i18n) support

hatesalp:
## Description

- Integrated react-i18next with bundled JSON resources for `tr` and `en`
- Added i18n configuration with `common` namespace and Turkish as the fallback language
- Replaced all hardcoded strings in Upload components with `t()` translation calls
- Implemented runtime language switching via `setLocale` in `MotifProvider`
- Synced `i18n.changeLanguage` with the locale state using `useEffect`

### Updated Components
- MotifProvider (handles locale state + i18n sync)
- UploadInput, UploadDragger, UploadList, ImageUpload (migrated hardcoded strings → `t()` usage)
- UploadProvider (translation-aware error messages)

### Runtime flow
Components use `useTranslation()` from react-i18next to access the `t()` function.

Translation keys (e.g. `upload.message.maxSizeError`) are resolved from JSON files located at: src/lib/i18n/locales/{lang}/common.json

MotifProvider keeps the current locale in state and syncs it with `i18n.changeLanguage()` via `useEffect`. 
When the language changes, all translated content updates automatically.

---

### Test (Jest) flow
In tests, react-i18next is fully mocked in `jest.setup.js`, so the real i18n runtime is never initialized.

Instead, a custom `mockT` function mimics the core behavior of `t()`:

- Resolves nested keys like `upload.message.maxSizeError` from the Turkish JSON
- Supports interpolation (e.g. `{{maxSize}}` → "5MB")

This `mockT` is injected into `useTranslation()` via `jest.mock`, so components receive it just like in production.

It’s also exposed on `globalThis`, allowing tests to directly call:
mockT(MESSAGE.UPLOAD_SUCCESS)

---

Added `@rollup/plugin-json` to support including JSON translation files (src/i18n/locales/{tr,en}/common.json) in the library package.

Excluded `/testHelper.*` from the rollup TypeScript build, so test-only helpers based on the global `mockT` no longer break the production build.

### Why this approach?
- `useTranslation()` is a hook, so it can't be used at the top level in test files
- Mocking avoids dealing with the i18n initialization lifecycle
- Both components and test assertions use the same `mockT`, ensuring consistent results
## Type of Change

- [x] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Screenshots / Preview

{52BE8E24-04C9-4EB0-AF4F-6C4B93B1FD69}

{F280A1D5-D936-4BC3-84E8-EE4AE3CA7BBC}

## Checklist

## How to Test

Closes [#EDKUI-1256]: Internal purposes only 
Closes [#EDKUI-1257]: Internal purposes only 
Closes [#EDKUI-1264]: Internal purposes only


aktasmehmet:
I've added some commits for refinement. Here are the updates:

- Removed react-i18next from the library — UI libraries shouldn't own a global i18n singleton; it conflicts with the consuming app's own i18n setup.                                                    
  - Built a zero-dependency translator (src/i18n/translate.ts) — Simple createTranslator(locale) factory with dot-notation key lookup, {{param}} interpolation, and three-layer fallback (override texts →
   locale JSON → English → raw key).                                                                                                                                                                      
  - Auto-derived Locale type and LocaleKey autocomplete from JSON (src/i18n/types.ts) — No hardcoded string unions; adding a new locale file automatically updates the type and IDE autocomplete.
  - locales/en/common.json → locales/en.json, same for tr; 
  - single registration point in locales/index.ts.                                                               
  - locale prop accepts a plain string or an object ({ locale?: Locale; texts?: DeepPartial<LocaleShape> }) — Allows runtime locale switching                                                                                                                                                                         
  - External t prop on MotifProvider — Consumers using react-i18next or any other library can pass their own t and the library defers to it entirely.                                                     
  - Migrated all Upload components off useTranslation — They now read t from useMotifContext(). 
  - Fixed a missed hardcoded "Yükle" in UploadButtonArea.tsx.                                                 
  - Replaced the custom mockT test setup — Removed the duplicated translator logic and react-i18next mock from jest.setup.js; t = createTranslator("en") now lives in src/utils/testUtils.ts as a shared, 
  typed export for all test files.

aktasmehmet (dismissed):

aktasmehmet (dismissed):

aktasmehmet (dismissed):

aktasmehmet (approved):

atakanbayrak (approved):

PR #48, opened 2026-04-16
fix: add gutter prop to Grid

ZehranurC:
## Description

- Added `gutter` prop to Grid (`xs | sm | md | lg`), defaults to `md`
- Design token file updated with new gutter values aligned with the design team

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Checklist

## How to Test

```
// xs gutter
<Grid gutter="xs">
  <Grid.Row>
    <Grid.Col>Col 1</Grid.Col>
    <Grid.Col>Col 2</Grid.Col>
  </Grid.Row>
</Grid>

// sm gutter
<Grid gutter="sm">
  <Grid.Row>
    <Grid.Col>Col 1</Grid.Col>
    <Grid.Col>Col 2</Grid.Col>
  </Grid.Row>
</Grid>

// md gutter
<Grid gutter="md">
  <Grid.Row>
    <Grid.Col>Col 1</Grid.Col>
    <Grid.Col>Col 2</Grid.Col>
  </Grid.Row>
</Grid>

// lg gutter
<Grid gutter="lg">
  <Grid.Row>
    <Grid.Col>Col 1</Grid.Col>
    <Grid.Col>Col 2</Grid.Col>
  </Grid.Row>
</Grid>
```
Closes [#EDKUI-1351]


src/lib/components/Grid/Grid.tsx
```diff
@@ -8,13 +8,16 @@ import { GridProps } from "./types";
...
 import usePropsWithThemeDefaults from "../../motif/hooks/usePropsWithThemeDefaults";
 import GridContext from "@/components/Grid/GridContext";
+import { useContext } from "react";
 
 const GridComponent = (props: PropsWithRef<GridProps, HTMLDivElement>) => {
   const { fluid, leanToEdge, children, className, style, colProps = {}, ref } = usePropsWithThemeDefaults("Grid", props);
-  const classNames = sanitizeModuleRootClasses(styles, className, [fluid && "fluid", leanToEdge && "leanToEdge"]);
+  const { isNested } = useContext(GridContext);
+  const resolvedLeanToEdge = Boolean(leanToEdge ?? isNested);
+  const classNames = sanitizeModuleRootClasses(styles, className, [fluid && "fluid", resolvedLeanToEdge && "leanToEdge"]);
```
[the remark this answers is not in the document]
ZehranurC: This section has already been reversed.

src/lib/styles/themes/default-theme.css
```diff
@@ -8,6 +8,9 @@
 :root {
   --theme-sizing-limit-input-min-width: 10rem;
   --theme-grid-gutter-gutter: 1.5rem;
+  --theme-grid-gutter-gutter-md: 1rem;
```
aktasmehmet: those values should also be included into the papyrus theme

src/lib/components/Grid/Grid.test.tsx
```diff
@@ -15,6 +17,17 @@ describe("Grid", () => {
     ).toMatchSnapshot();
   });
 
+  it.each(gutterSizes)("should apply gutter-%s class when gutter prop is %s", size => {
```
aktasmehmet: this test should be seperated. This is a basic prop test we do all the time

src/lib/components/Grid/Grid.module.scss
```diff
@@ -12,8 +18,13 @@ $grids: 12;
...
   flex-direction: column;
   box-sizing: border-box;
-  gap: var(--theme-grid-gutter-gutter);
-  padding: var(--theme-grid-gutter-gutter);
+
+  @each $size, $token in $gutter-token {
+    &.gutter-#{$size} {
+      gap: var(--theme-grid-gutter-#{$token});
+      padding: var(--theme-grid-gutter-#{$token});
+    }
```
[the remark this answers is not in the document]
aktasmehmet: I was gonna say the same thing. Row is not effected.

Besides, you can but these loop outside of the root without using "&" at the beginning. This would make the classes less spesific for possible future overrides.

src/lib/components/Grid/Grid.module.scss
```diff
@@ -1,6 +1,12 @@
...
 $grid-sizes: ("sm", "md", "lg", "xl");
+$gutter-token: (
+  "xs": "gutter-xs",
+  "sm": "gutter-sm",
+  "md": "gutter-md",
+  "lg": "gutter"
+);
 $grids: 12;
 @function calculate-grid-flex-width($colIdx) {
   @return calc(($colIdx / $grids * 100%) - ((1 - $colIdx / $grids) * var(--theme-grid-gutter-gutter)));
```
aktasmehmet: calculate-grid-flex-width uses default gutter. This would probably change with the new gutter sizes

aktasmehmet (changes requested):

src/lib/styles/themes/default-theme.css
hatesalp: +1

src/lib/components/Grid/Grid.module.scss
```diff
@@ -1,9 +1,21 @@
 @use "@styles" as *;
 
 $grid-sizes: ("sm", "md", "lg", "xl");
+$gutter-token: (
+  "xs": "gutter-xs",
+  "sm": "gutter-sm",
+  "md": "gutter-md",
+  "lg": "gutter"
+);
```
aktasmehmet: I'm gonna state every issue here:

- 1 - There is an unused ``--theme-grid-gutter-gutter`` token now. From now on, there is no default gutter. So, the tokens should be:

```
--theme-grid-gutter-gutter-lg
--theme-grid-gutter-gutter-md
--theme-grid-gutter-gutter-sm
--theme-grid-gutter-gutter-xs
```

- 2 - So, gutter token map is not necessary. You may use just the keys. In fact, I suggest changing the names from xs-lg to sm-xl like the grid size to keep the consistency with less code. If you do so, you also have to change the token names. **Please dont forget to keep aligned with the design team and the Figma side**

- 3 - ``--grid-gutter`` is not necessary. You can iterate nested 2 loops or if you keep the same sizing name convention, with a single loop, you can give the necessary value.

- 4 - So, instead of setting like ``gap: var(--grid-gutter);`` in Root, you have to set something like:

```
@each $size in $sizes {
  &.gutter-#{$size} {
    gap: var(--theme-grid-gutter-#{$size});
    padding: var(--theme-grid-gutter-#{$size});
  }
}
```

- 5 - instead of using ``var(--grid-gutter)`` in ``calculate-grid-flex-width`` in you can pass the gutter value to the function itself as a parameter.

aktasmehmet (changes requested):

src/lib/components/Grid/Grid.module.scss
ZehranurC: The &.gutter-{size} pattern was originally nested inside .Root in a previous commit (319d64a), but was moved outside following Copilot and your suggestions. In the last commit, I updated the size naming, moved it back inside Root, and applied the other updates as well.

aktasmehmet (approved):

hatesalp (approved):

PR #50, opened 2026-04-28
test: add Timeline component tests

ZehranurC:
## Description

- Timeline tests have been written

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Screenshots / Preview
image

## Checklist

## How to Test

Closes [#EDKUI-1330]


src/lib/components/Timeline/Timeline.test.tsx
```diff
@@ -0,0 +1,183 @@
...
+const items: TimelineItemProps[] = [
+  { title: "Step 1", content: "First step content" },
+  { title: "Step 2", content: "Second step content" },
+  { title: "Step 3", content: "Third step content" },
+];
+
+describe("Timeline", () => {
+  it("should be rendered with only required props and should have default prop values stated here", () => {
+    const { container } = render(<Timeline items={items} />);
+    expect(container).toMatchSnapshot();
```
atakanbayrak, line 15: This line is duplicated within many tests.

src/lib/components/Timeline/Timeline.test.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+    // textAlign: start
+    expect(root).toHaveClass("start");
+
+    // variant: primary
+    expect(root).toHaveClass("primary");
+  });
+
+  it("should be rendered as given in orientation prop", () => {
+    const { container, rerender } = render(<Timeline items={items} orientation="vertical" />);
```
atakanbayrak: I guess vertical orientation prop is default so maybe no need to state value here.

src/lib/components/Timeline/Timeline.test.tsx
```diff
@@ -0,0 +1,183 @@
...
+    const root = container.firstChild as HTMLElement;
+    expect(root).toHaveClass("vertical");
+
+    rerender(<Timeline items={items} orientation="horizontal" />);
+    expect(root).toHaveClass("horizontal");
+  });
+
+  it("should be rendered as given in contentPosition prop", () => {
+    const { container, rerender } = render(<Timeline items={items} contentPosition="after" />);
```
atakanbayrak: Same as above, no need to state contentPosition prop here I guess.

src/lib/components/Timeline/Timeline.test.tsx
```diff
@@ -0,0 +1,183 @@
...
+    render(<Timeline items={[{ title: "My Title" }]} />);
+    expect(screen.getByText("My Title")).toBeInTheDocument();
+  });
+
+  it("should render item content text", () => {
+    render(<Timeline items={[{ content: "My Content" }]} />);
+    expect(screen.getByText("My Content")).toBeInTheDocument();
+  });
+
+  it("should render both title and content when both are provided", () => {
```
atakanbayrak: Usage of "both" is duplicated inside test naming sentence.

src/lib/components/Timeline/Timeline.test.tsx
```diff
@@ -0,0 +1,183 @@
...
+    expect(container.querySelector(".content")).toBeInTheDocument();
+    expect(container.querySelector(".title")).not.toBeInTheDocument();
+  });
+
+  it("should not render section when item has neither title nor content", () => {
+    const { container } = render(<Timeline items={[{}]} />);
+    expect(container.querySelector(".section")).not.toBeInTheDocument();
+  });
+
+  it("should render number markers with correct incremental order when markerType is number", () => {
```
atakanbayrak, line 114: This checking methodology does not provide that they are in the correct order, it could still look like 2 - 1 - 3 if there is an error on the logic. Maybe we can look deeper item components to check if there is correct number.

src/lib/components/Timeline/Timeline.test.tsx
```diff
@@ -0,0 +1,183 @@
...
+  it("should apply outlined appearance class when appearance prop is outlined", () => {
+    const { container } = render(<Timeline items={[{ title: "Item", appearance: "outlined" }]} />);
+    expect(container.querySelector(".item")).toHaveClass("outlined");
+  });
+
+  it("should apply item-level variant class only on the item that has variant prop", () => {
+    const { container } = render(
+      <Timeline items={[{ title: "Default item" }, { title: "Success item", variant: "success" }]} variant="primary" />,
+    );
+    const itemEls = container.querySelectorAll(".item");
```
atakanbayrak: Normally I don't write down this const naming but there will be a new commit either so we could change this naming to better one.

src/lib/components/Timeline/Timeline.test.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+  it("should apply custom className to root element", () => {
+    const { container } = render(<Timeline items={items} className="custom-class" />);
+    expect(container.firstChild).toHaveClass("custom-class");
+  });
+
+  it("should apply custom style to root element", () => {
+    const { container } = render(<Timeline items={items} style={{ width: "500px" }} />);
+    expect(container.firstChild).toHaveStyle({ width: "500px" });
+  });
```
atakanbayrak: I guess these cases are not needed.

atakanbayrak (changes requested):

src/lib/components/Timeline/Timeline.test.tsx
```diff
@@ -0,0 +1,183 @@
...
+});
```
hatesalp, line 157: A test should be added to display the default "motif_ui" icon when markerType="icon" is not provided.

hatesalp (dismissed):
lgtm 💮

src/lib/components/Timeline/Timeline.test.tsx
aktasmehmet: +1

src/lib/components/Timeline/Timeline.test.tsx
aktasmehmet: it looks like it is added. resolving

src/lib/components/Timeline/Timeline.test.tsx
```diff
@@ -0,0 +1,179 @@
...
+  it("should apply item-level variant for all item variant values", () => {
+    const variants: TimelineVariant[] = ["primary", "secondary", "success", "warning", "danger", "light"];
+    variants.forEach(variant => {
+      const { container, unmount } = render(<Timeline items={[{ title: "Item", variant }]} />);
+      expect(container.querySelector(".item")).toHaveClass(variant);
+      unmount();
+    });
+  });
+
+  it("should not render any items when items array is empty", () => {
```
aktasmehmet: This test is redundant because ``should render the correct number of items`` already covers this.

src/lib/components/Timeline/Timeline.test.tsx
```diff
@@ -0,0 +1,179 @@
...
+  it("should apply item-level variant class only on the item that has variant prop", () => {
+    const { container } = render(
+      <Timeline items={[{ title: "Default item" }, { title: "Success item", variant: "success" }]} variant="primary" />,
+    );
+    const itemElements = container.querySelectorAll(".item");
+    expect(itemElements[0]).not.toHaveClass("success");
+    expect(itemElements[1]).toHaveClass("success");
+  });
+
+  it("should apply item-level variant for all item variant values", () => {
```
aktasmehmet:
```suggestion
  it("should render item in the given variant when variant prop in Timeline itself is not set", () => {
```

src/lib/components/Timeline/Timeline.test.tsx
```diff
@@ -0,0 +1,179 @@
...
+    const { container } = render(<Timeline items={[{ title: "Item" }]} />);
+    expect(container.querySelector(".item")).toHaveClass("filled");
+  });
+
+  it("should apply outlined appearance class when appearance prop is outlined", () => {
+    const { container } = render(<Timeline items={[{ title: "Item", appearance: "outlined" }]} />);
+    expect(container.querySelector(".item")).toHaveClass("outlined");
+  });
+
+  it("should apply item-level variant class only on the item that has variant prop", () => {
```
aktasmehmet: This test should cover whether the component level variant overrides item level variant. Other tests are already covered.

Besides, giving technical details in test is not a good approach, like class name, etc. We only care about the rendered end result.

So:

- 1 item with variant is enough
- You should test that item has not "success", but primary (or whatever class names are applied to see the primary in browser)
- Test name should also be updated as ``should render the variant given in Timeline component when both Timeline variant and item variant is provided``

src/lib/components/Timeline/Timeline.test.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+  it("should apply filled appearance class by default on items", () => {
+    const { container } = render(<Timeline items={[{ title: "Item" }]} />);
+    expect(container.querySelector(".item")).toHaveClass("filled");
+  });
+
+  it("should apply outlined appearance class when appearance prop is outlined", () => {
+    const { container } = render(<Timeline items={[{ title: "Item", appearance: "outlined" }]} />);
+    expect(container.querySelector(".item")).toHaveClass("outlined");
+  });
```
aktasmehmet: "default filled" should be moved into the first test as we always do.

And ``should apply outlined appearance class when appearance prop is outlined`` test should be renamed as ``render the items in appearance given in appearance prop of the item``

src/lib/components/Timeline/Timeline.test.tsx
```diff
@@ -0,0 +1,179 @@
...
+    render(<Timeline items={[{ title: "Item" }]} markerType="icon" />);
+    expect(screen.getByText("motif_ui")).toBeInTheDocument();
+  });
+
+  it("should render custom icon name when markerType is icon and icon prop is set", () => {
+    render(<Timeline items={[{ title: "Item", icon: "check" }]} markerType="icon" />);
+    expect(screen.getByText("check")).toBeInTheDocument();
+  });
+
+  it("should apply disabled class on item when disabled prop is true", () => {
```
aktasmehmet:
```suggestion
  it("should render the item with disabled is true, as disabled", () => {
```

src/lib/components/Timeline/Timeline.test.tsx
```diff
@@ -0,0 +1,179 @@
...
+    expect(itemElements[1].querySelector(".markerNumber")).toHaveTextContent("2");
+    expect(itemElements[2].querySelector(".markerNumber")).toHaveTextContent("3");
+  });
+
+  it("should render default motif_ui icon when markerType is icon and no icon prop is set", () => {
+    render(<Timeline items={[{ title: "Item" }]} markerType="icon" />);
+    expect(screen.getByText("motif_ui")).toBeInTheDocument();
+  });
+
+  it("should render custom icon name when markerType is icon and icon prop is set", () => {
```
aktasmehmet:
```suggestion
  it("should render the icons in icon prop of the items only when markerType is icon", () => {
```

More important is that this test should also cover that the icons are not rendered when markerType is not icon

src/lib/components/Timeline/Timeline.test.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+  it("should render number markers with correct incremental order when markerType is number", () => {
+    const { container } = render(<Timeline items={items} markerType="number" />);
+    const itemElements = container.querySelectorAll(".item");
+    expect(itemElements[0].querySelector(".markerNumber")).toHaveTextContent("1");
+    expect(itemElements[1].querySelector(".markerNumber")).toHaveTextContent("2");
+    expect(itemElements[2].querySelector(".markerNumber")).toHaveTextContent("3");
+  });
+
+  it("should render default motif_ui icon when markerType is icon and no icon prop is set", () => {
```
aktasmehmet: This test should move into the first one since it has a default value check in it.

src/lib/components/Timeline/Timeline.test.tsx
```diff
@@ -0,0 +1,179 @@
...
+    const { container } = render(<Timeline items={[{}]} />);
+    expect(container.querySelector(".section")).not.toBeInTheDocument();
+  });
+
+  it("should render number markers with correct incremental order when markerType is number", () => {
+    const { container } = render(<Timeline items={items} markerType="number" />);
+    const itemElements = container.querySelectorAll(".item");
+    expect(itemElements[0].querySelector(".markerNumber")).toHaveTextContent("1");
+    expect(itemElements[1].querySelector(".markerNumber")).toHaveTextContent("2");
+    expect(itemElements[2].querySelector(".markerNumber")).toHaveTextContent("3");
```
aktasmehmet: Maybe you should think to convert them a single liner, without duplication?

```
container.querySelectorAll(".item")
  .forEach((item, idx) => 
    expect(item.querySelector(".markerNumber"))
    .toHaveTextContent(`${idx + 1}`));
```

src/lib/components/Timeline/Timeline.test.tsx
```diff
@@ -0,0 +1,179 @@
...
+  it("should not render title span when only content is provided", () => {
+    const { container } = render(<Timeline items={[{ content: "My Content" }]} />);
+    expect(container.querySelector(".content")).toBeInTheDocument();
+    expect(container.querySelector(".title")).not.toBeInTheDocument();
+  });
+
+  it("should not render section when item has neither title nor content", () => {
+    const { container } = render(<Timeline items={[{}]} />);
+    expect(container.querySelector(".section")).not.toBeInTheDocument();
+  });
```
aktasmehmet: other "should render" tests are already covering all these 3 tests. I think there are unnecessary and can be deleted.

src/lib/components/Timeline/Timeline.test.tsx
```diff
@@ -0,0 +1,179 @@
...
+    markerTypes.forEach(markerType => {
+      const { container, unmount } = render(<Timeline items={items} markerType={markerType} />);
+      expect(container.firstChild).toHaveClass(markerType);
+      unmount();
+    });
+  });
+
+  it("should render the correct number of items", () => {
+    const { container } = render(<Timeline items={items} />);
+    expect(container.firstChild?.childNodes.length).toBe(items.length);
```
aktasmehmet: I think checking "Step 1", "Step 2" and "Step 3" would be a better approach instead of just checking the number of items. Because as far as I see, there is no simple test that checks "given items are rendered". So, in this test you can also cover that too. Step1-2-3 availability also covers the number of items as well.

src/lib/components/Timeline/Timeline.test.tsx
```diff
@@ -0,0 +1,179 @@
...
+  it("should be rendered in the variant given in the variant prop", () => {
+    const variants: TimelineVariant[] = ["primary", "secondary", "success", "warning", "danger", "light"];
+    variants.forEach(variant => {
+      const { container, unmount } = render(<Timeline items={items} variant={variant} />);
+      expect(container.firstChild).toHaveClass(variant);
+      unmount();
+    });
+  });
+
+  it("should be rendered as given in markerType prop", () => {
```
aktasmehmet:
```suggestion
  it("should render the markers in appearance as given in markerType prop", () => {
```

src/lib/components/Timeline/Timeline.test.tsx
```diff
@@ -0,0 +1,179 @@
...
+    expect(root).toHaveClass("start");
+
+    rerender(<Timeline items={items} textAlign="center" />);
+    expect(root).toHaveClass("center");
+
+    rerender(<Timeline items={items} textAlign="end" />);
+    expect(root).toHaveClass("end");
+  });
+
+  it("should be rendered in the variant given in the variant prop", () => {
```
aktasmehmet:
```suggestion
  it("should render in the given variant", () => {
```

src/lib/components/Timeline/Timeline.test.tsx
```diff
@@ -0,0 +1,179 @@
...
+    expect(root).toHaveClass("after");
+
+    rerender(<Timeline items={items} contentPosition="before" />);
+    expect(root).toHaveClass("before");
+
+    rerender(<Timeline items={items} contentPosition="alternate" />);
+    expect(root).toHaveClass("alternate");
+  });
+
+  it("should be rendered as given in textAlign prop", () => {
```
aktasmehmet:
```suggestion
  it("should render the content aligned as given in textAlign prop", () => {
```

src/lib/components/Timeline/Timeline.test.tsx
```diff
@@ -0,0 +1,179 @@
...
+  it("should be rendered as given in orientation prop", () => {
+    const { container, rerender } = render(<Timeline items={items} />);
+    const root = container.firstChild as HTMLElement;
+    expect(root).toHaveClass("vertical");
+
+    rerender(<Timeline items={items} orientation="horizontal" />);
+    expect(root).toHaveClass("horizontal");
+  });
+
+  it("should be rendered as given in contentPosition prop", () => {
```
aktasmehmet:
```suggestion
  it("should position the content for each item as given in contentPosition prop", () => {
```

src/lib/components/Timeline/Timeline.test.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+    // markerType: dot
+    expect(root).toHaveClass("dot");
+
+    // textAlign: start
+    expect(root).toHaveClass("start");
+
+    // variant: primary
+    expect(root).toHaveClass("primary");
+  });
```
aktasmehmet, line 43: you should add the defaults for the items as well (I've also mentioned in each test of them)

aktasmehmet (changes requested):

src/lib/components/Timeline/Timeline.test.tsx
```diff
@@ -0,0 +1,156 @@
...
+    const { container, rerender } = render(<Timeline items={[{ title: "Item", appearance: "filled" }]} />);
+    expect(container.querySelector(".item")).toHaveClass("filled");
+
+    rerender(<Timeline items={[{ title: "Item", appearance: "outlined" }]} />);
+    expect(container.querySelector(".item")).toHaveClass("outlined");
+  });
+
+  it("should render the variant given in item when both Timeline variant and item variant are provided", () => {
+    const { container } = render(<Timeline items={[{ title: "Item", variant: "success" }]} variant="primary" />);
+    expect(container.querySelector(".item")).toHaveClass("success");
```
aktasmehmet, line 145: We should also check that:

```expect(container.querySelector(".item")).not.toHaveClass("primary");```

aktasmehmet (commented):
It's perfect! Just one small request and it's good to go 🙏

aktasmehmet (approved):

atakanbayrak (approved):

PR #51, opened 2026-04-29
fix: internal inputText component is added

aktasmehmet:
## Description

A new InputText component for internal purposes is created. This new input provides a common wrapper where an input text is needed. It has lots of features that are possibly needed. 

Here are the notable changes:

- /motif/InputText is created.
- Current InputText now completely uses the new one.
- In order to keep mtf-InputText class available, same component name is used.
- value keep and change mechanism is updated. Ref objects are being used now.
- Rendered elements are completely re-coded along with the new scss file.
- Necessary new tests are added and previous tests are updated (mostly snapshots).
- Commond tests are not included on purpose. This component is only a helper and the features should be tested bby the components that use this component. 

**New Props:**
- ``disableTyping``: this is now being used instead of ``readOnlyWithEnabledLook``. It disables typing while keeping everything else active.
- ``uncontrolled``: makes the component uncontrolled
- ``buttonRight``: adds a new button to the right side inside the input for purposes like toggling, etc.
- ``clearable``: it creates the necessary icon button to reset the value.
- ``onKeyUp``: callback for keyUp event.
- ``type``: sets the type of the input. Values are, text, number, password. (Internal Input Text)
- ``type``: sets the type of the input. Values are, text, number. (Original Input Text)

**Removed Props:**
- ``readOnlyWithEnabledLook``

**OTHER:**

- Sanitize function in css utils is updated to eliminate duplicate class names.
- Some components without readOnly prop are not fixed.
- Some updates from Figma design are reflected to ``Dropdown``. 
- useRegisterFormField, did not expose onChange event properly. This is fixed and a new function is added for the registered items(onChangeWithFormFieldUpdate). By doing that, there is no need to call ``onFormFieldValueUpdate`` in onchange of the inputs

## Type of Change

- [x] 🆕 New Component
- [x] ✨ Feature / Enhancement
- [x] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [x] 🧹 Code Refactoring

## Screenshots / Preview

Screenshot 2026-04-29 at 14 30 42

## Checklist

## How to Test

- InputText or /motif/Input text components can be tested. They should render %100 identical view.
- These component are highly coupled with Form component. So, a strong form related testing is needed. 

_Closes [#EDKUI-818]: Internal purposes only_


src/lib/components/InputText/InputText.test.tsx
```diff
@@ -100,4 +100,14 @@ describe("InputText", () => {
     const { getByTestId } = render(<InputText pill />);
     expect(getByTestId("inputItem")).toHaveClass("pill");
   });
+
+  it("should render the input in given type", () => {
```
atakanbayrak: This test could be added;

- When none number value is written down, increment and decrement icons reset it and starts from 0 so we should not loose this feature.

src/lib/components/Motif/InputText/InputText.module.scss
```diff
@@ -0,0 +1,129 @@
...
+
+  .icon,
+  input {
+    color: var(--theme-color-text-base-disabled);
+  }
+  .numberButtons button {
+    color: var(--theme-color-text-base-disabled);
+    background-color: var(--theme-color-surface-base-base3);
+  }
+}
```
atakanbayrak, line 86: readOnly and disabled behaving like they are same, we have talked about this situation before but I am still kinda confused about it.

atakanbayrak (changes requested):

src/lib/components/InputDateRange/InputDateRange.tsx
```diff
@@ -131,12 +131,12 @@ const InputDateRange = (p: PropsWithRef<InputDateRangeProps, HTMLDivElement>) =>
...
         disabled={disabled}
-        readOnlyWithEnabledLook
+        disableTyping
         value={typedValue}
         onClick={pickerShowHandler}
         onFocus={pickerShowHandler}
         iconLeft={<MotifIcon name="calendar_expand_horizontal" size={size} />}
-        iconRight={itemValue && !disabled && <MotifIconButton name="cancel_outline" size={size} onClick={onClearClickInInput} />}
+        iconRight={itemValue && !disabled ? <MotifIconButton name="cancel_outline" size={size} onClick={onClearClickInInput} /> : undefined}
       />
```
hatesalp, line 140: Why not use the clearable prop instead of passing a MotifIconButton through iconRight?
The only functional difference I can see is that the current approach conditionally hides the button when itemValue is empty, while clearable always renders it. If that's the blocker, we could update InternalInputText to hide the clear button when the value is empty. Which arguably makes more sense as a default behavior anyway.

src/lib/components/Motif/InputText/InputText.module.scss
aktasmehmet: they are the same "in terms of UI". But the attributes applied to the input is different. When it is disabled, disabled attr is given, when readonly, then readonly. This decides the form submit behavior. Other than that, their UI is the same.

I wanted to keep the class names seperate for just class name convention standards.

We also have another option to look like it is not readonly but the input is readonly. In that scenario, the UI looks normal so this block is not applied there.

src/lib/components/InputDateRange/InputDateRange.tsx
aktasmehmet: You are absolutely right but I was gonna create a seperate task to update this type of components to update. I dont want to merge these 2 tasks together. Probably, there are more of this kind of code to update other than just InputDateRange.

src/lib/components/InputText/InputText.test.tsx
aktasmehmet: 2 things:

1 - It is impossible to put a non number value into an input text with type="number"
2 - It is only possible for an empty string and that "starts from 0" behavior is the behevior of the html <input /> it self. So we dont need to check that here, again.

But, I've forgot to check the increment, decrement behavior and just put that part into the test. Thx 🙏

src/lib/components/InputDateRange/InputDateRange.tsx
hatesalp: Okay, I have nothing else to add. It looks very good. :clap:  I approve.

hatesalp (approved):

atakanbayrak (approved):

hatesalp (approved):

PR #53, opened 2026-04-30
fix: min-width for form is removed

atakanbayrak:
## Description

min-width css value under Form.module.css which causes error to define exact amount of width is deleted, error correction is checked with the installation of fixed library version inside motif react website.

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [x] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Checklist

## How to Test

#EDKUI-1375

### DIFF
--- a/src/lib/components/Form/Form.module.scss
+++ b/src/lib/components/Form/Form.module.scss
@@ -9,7 +9,6 @@ $sizes: (
 );
 
 .Root {
-  min-width: min-content;
   display: flex;
   flex-direction: column;
 }
--- a/src/lib/components/Form/Form.module.scss
+++ b/src/lib/components/Form/Form.module.scss
@@ -17,6 +17,10 @@ $sizes: (
   display: flex;
 }
 
+.horizontal {
+  display: inline-flex;
+}
+
 .horizontal .fields {
   display: inline-flex;
   align-items: flex-start;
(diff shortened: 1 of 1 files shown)


src/lib/components/Form/Form.module.scss
```diff
@@ -9,7 +9,6 @@ $sizes: (
 );
 
 .Root {
-  min-width: min-content;
   display: flex;
```
hatesalp: I think we shouldn’t remove min-width: min-content. In the horizontal layout, .fields uses inline-flex, so the children line up side by side and can overflow .Root. This property helps .Root grow with its content and prevents that overflow.

If needed, we can look for an alternative solution, but removing it might cause the problem to reappear.

Image

aktasmehmet: Alternative solution would be a better approach. The easiest way is:

1 - Keep ```min-width: content``` removed 
2 - Make .Root ```display:inline-flex``` when form orientation is horizontal.

This would solve the issue I guess

src/lib/components/Form/Form.module.scss
```diff
@@ -18,6 +17,10 @@ $sizes: (
...
```
aktasmehmet, line 17: C'mon, why keep this here? It is only valid for vertical orientation so it should be moved under .vertical class, like you did for horizontal

aktasmehmet (changes requested):

aktasmehmet (approved):

hatesalp (approved):

PR #55, opened 2026-04-30
fix: accordion form submit issue and improve group expand sync

hatesalp:
## Description

- The issue where the accordion triggered form submission when used inside a form was fixed by setting the header button type="button"
- The incorrect initial open state in single-expand AccordionGroup was fixed (now expandedIndex from the group is used instead of expanded)
- State synchronization on expanded prop changes was improved by managing it through setExpandedIndex
- The end state where all accordions were collapsed (expanded Index === -1) was reopened by setting the expanded value of the accordion to true.
 - Snapshot files have been updated.

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [x] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Checklist

## How to Test
```
    <Accordion.Group multiExpand={false} condensed>
        <Accordion index={0} icon="person" title="Accordion Title 1" text="First accordion content." />
        <Accordion index={1} expanded icon="settings" title="Accordion Title 2" text="Second accordion content." />
        <Accordion index={2} expanded icon="info" title="Accordion Title 3" text="Third accordion content." />
      </Accordion.Group>
```

```
 <Form onSubmit={e => { console.log(e); }}>
      <Accordion.Group multiExpand condensed>
          <Accordion expanded index={0} icon="person" title="Accordion Title 1" text="First accordion content." />
          <Accordion expanded index={1} icon="settings" title="Accordion Title 2" text="Second accordion content." />
          <Accordion expanded index={2} icon="info" title="Accordion Title 3" text="Third accordion content." />
      </Accordion.Group>
 </Form>

```

 [#EDKUI-1396]:


atakanbayrak (dismissed):
Wise solution I guess, thanks for the improvement.

src/lib/components/Accordion/Accordion.tsx
```diff
@@ -24,21 +24,27 @@ const AccordionComponent = (props: PropsWithRefAndChildren<AccordionProps, HTMLD
     className,
   } = usePropsWithThemeDefaults("Accordion", props);
   const { expandedIndex, setExpandedIndex, multiExpand, groupEnabled } = useContext(AccordionGroupContext);
-  const { visible: open, toggle } = useToggle(expanded);
+  const initialOpen = groupEnabled && !multiExpand ? expandedIndex === index : expanded;
```
aktasmehmet: This is not a good approach. Because when there are multiple expanded Accordions, their initialOpen will be true both. So, the useEffect with expanded dependency will clear the mess. This basically shows that both accordions will be open for a second and then one of them will collapse. This is not a good UX approach.

Instead, we should directly show the expanded, the only expanded Accordion.

I propose a solution to this. I'll send a commit in a second and you can check. Basically just cheking the expanded props, before they are sent to the context. So, if there are illegal multiple expanded props, only one of them is selected as the expandedIndex in Group context. So the issue is fixed. I am also adding some helper consts to the Accordion itself.

aktasmehmet (changes requested):
I'd also love to see a test to cover this issue

src/lib/components/Accordion/__snapshots__/Accordion.test.tsx.snap
```diff
@@ -8,6 +8,7 @@ exports[`Accordion should be rendered with only required props 1`] = `
   >
```
aktasmehmet, line 8: I'd also love to see a test to cover the whole issue

aktasmehmet: I've just added 😉

aktasmehmet (approved):

atakanbayrak (approved):

PR #57, opened 2026-04-30
fix: flex wrap is added for pagination

atakanbayrak:
## Description

Problems which occurs when using PinCode and Pagination solved with flex-wrap prop inside css.

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [x] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Checklist

## How to Test

#EDKUI-1385

### DIFF
--- a/src/lib/components/Pagination/Pagination.module.scss
+++ b/src/lib/components/Pagination/Pagination.module.scss
@@ -9,6 +9,7 @@ $pagination-sizes: (
 
 .Root {
   display: inline-flex;
+  flex-wrap: wrap;
 }
 
 .pageItem {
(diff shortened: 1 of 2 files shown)


ZehranurC (dismissed):

aktasmehmet (dismissed):

aktasmehmet (changes requested):
For the pagination, it is fine.

For PinCode, we should keep it as nowrap. Because we are having issues when using form in horizontal view. Besides, the best practices also suggest to keep it as block without wrap. So I guess you should revert only the PinCode part.

Image

aktasmehmet (approved):

aktasmehmet:
Can you please update the PR title? It is only for the Pagination now.

hatesalp (approved):

PR #61, opened 2026-05-04
fix: useState logic is introduced for avoiding bug inside tooltip caret hovering

atakanbayrak:
## Description

I introduced both useState and useRef solution for usePositionTooltip.

`lastTriedPosition` was tracked only with `useRef`. This worked for reading the latest value inside callbacks without stale closure issues, but it had a drawback: since ref mutations don't trigger re-renders, any consumer of `lastTriedPosition` outside this hook would never reflect the updated position.

To fix this, `useState` was introduced:

- **`lastTriedPositionRef` (useRef)** — used for reading the current value inside `useCallback` functions. Because these callbacks close over their dependencies at creation time, reading from a plain state variable would risk returning a stale value. The ref always holds the latest value regardless of when the callback was created.

- **`lastTriedPosition` (useState)** — used to propagate the position change to the outside world. Updating this triggers a re-render, ensuring any component consuming the returned `lastTriedPosition` always sees the up-to-date value.

Using only `useRef` would mean consumers never re-render on position changes. Using only `useState` would cause stale closure issues inside callbacks. The combination of both gives us synchronous, always-fresh reads inside the hook and reactive updates outside of it.

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [x] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Checklist

## How to Test

Solution is checkable from local storybook and localhost.

#EDKUI-1378


src/lib/components/Tooltip/usePositionTooltip.tsx
```diff
@@ -1,4 +1,4 @@
-import { CSSProperties, RefObject, useCallback, useRef, useState } from "react";
+import { CSSProperties, RefObject, useCallback, useRef, useState, useEffect } from "react";
```
aktasmehmet: I am writing my review here on top of the file.

There is really no need for this much change. If I am not missing anything, the solution is waaaay simpler.

The issue is that when it is first hovered, new position style is set, component is rerendered and the tooltip is positioned correctly. The issue is the caret tip as we know.

The problem here is that, since lastTriedPosition is a ref, it is not changed after rerender. You just need to set lastTriedPosition to position and set try counter to 0.

```
lastTriedPosition.current = position;
tryCounter.current = 0;  
```

Ttry to put this code in the very first lines of ``setTooltipPosition``. Thats it.

Can you revert the code, do this, and validate the solution?

aktasmehmet (changes requested):

src/lib/components/Tooltip/usePositionTooltip.tsx
atakanbayrak: I checked this solution both localhost and storybook and there is no problem mentioned on task.

aktasmehmet (approved):

hatesalp (approved):

PR #62, opened 2026-05-05
fix: add styling support to Tab.Panel wrappers

ZehranurC:
## Description

Added className and style props support to TabPanel. Custom styling is now possible.

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [x] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [x] 🧹 Code Refactoring

## Screenshots / Preview

image

## Checklist

## How to Test

Closes [#EDKUI-1376]


src/lib/components/Tab/types.ts
```diff
@@ -38,3 +39,7 @@ export type TabContextProps = {
   selectedTabId?: string;
   tabClickHandler?: (tabId: string) => void;
 };
+
+export type TabPanelProps = {
+  id: string;
+} & StandardProps;
```
aktasmehmet: TabPanelProps is used as a generic type in PropsWithChildren... Which already includes StandardProps. So, not necessary here.

aktasmehmet (changes requested):

src/lib/components/Tab/components/TabPanel.tsx
```diff
@@ -1,15 +1,20 @@
...
-};
-
-const TabPanel = (props: PropsWithChildren<Props>) => {
-  const { id, children } = props;
+const TabPanel = (props: PropsWithChildren<TabPanelProps>) => {
+  const { id, children, className, style } = props;
   const { selectedTabId } = useContext(TabContext);
 
-  return id === selectedTabId ? children : null;
+  const classNames = sanitizeModuleClasses(styles, "panelWrapper", className);
```
atakanbayrak: I didn't see that we have passed className prop inside sanitizeModuleClasses before. Does this approach legit to everyone ? I am not quite sure so I would like to have opinion of everyone. @aktasmehmet @hatesalp @ZehranurC

aktasmehmet: Nope, definitely missed that. Good catch. Now, it will search for styles[className] which will not be found of course. We should remove that. 

In fact, I've detected the same issue in ```DataViewItem.tsx```. @ZehranurC can you also fix that.

We did not applied class names from outside in sub components before. Thats why ``sanitizeModuleClasses``does not have a ``globalClass`` prop like in the ``sanitizeModuleRootClasses``

Right now, we need to append className to the classNames externally, not in the sanitize method.

ZehranurC: I've detected the same issue in the Col component and fixed it. I'll send it with the next commit.

aktasmehmet: I didnt like the solution. We already have 2 diffent uses to handle in the future. I was expecting another global sanitize function to handle all these.

I'm making that refactor in couple of minutes.

atakanbayrak (approved):

src/lib/components/Tab/components/TabPanel.tsx
ZehranurC: I didn't understand that you meant a sanitize method like this. This approach is much better. Well done!

aktasmehmet (approved):

PR #64, opened 2026-05-06
fix(inputDateRange): dateRangePicker input text name error when opening

aktasmehmet:
## Description

InputDateRange was throwing an error related to input text registration when the input is clicked and DateRangePicker is tried to be opened.

- The error in the description is fixed by replacing it with new internal text component
- Other minor refactor are mostly forgotten fixes and some of them are for the chromatic flaky test fixes.

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [x] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [x] 🧹 Code Refactoring

## Screenshots / Preview

BEFORE

Screenshot 2026-05-06 at 14 30 15

AFTER

Screenshot 2026-05-06 at 14 30 33

## Checklist

## How to Test

Closes [#EDKUI-1340]: Internal purposes only


src/docs/data/table.ts
atakanbayrak, line 4: We have this same method in here with minor changes but why did we create another one ? Did we specifically want all elements for each 'i' ?

atakanbayrak: Moreover, if main purpose is just change generateDummyTableData with generateMockTableData at the end of the day we can just delete generateDummyTableData because that will be never used anymore.

atakanbayrak (changes requested):

src/docs/components/FormOrientation/FormOrientation.stories.tsx
```diff
@@ -75,7 +75,7 @@ export const LabelOrientationComparison: Story = {
           <InputText iconLeft="person" />
         </Form.Field>
       </Form>
-      <Form onSubmit={() => {}} size="xs" labelOrientation="horizontal" style={{ width: 300 }} title="Horizontal">
```
atakanbayrak, line 78: https://www.chromatic.com/test?appId=651ec6529fd0775ff74f2cdf&id=69fc3372ea7ab8f15b5ba1e1

If nothing to add to comment, this issue could be resolved.

src/docs/data/table.ts
aktasmehmet: I just want to keep that function in case of a future use, especially for demo purposes.

The main reason I've added that function to get the same data all the time. This is for Chromatic. Since the table story always generates different data, that test is marked as flaky.

src/docs/components/FormOrientation/FormOrientation.stories.tsx
aktasmehmet: Yes, the main reason was to make this change. The previous one was problematic.

atakanbayrak (approved):

ZehranurC (approved):

PR #65, opened 2026-05-07
feat: css specificity has changed with @layer

atakanbayrak:
## Description

When consumers install this library and try to override component styles, they are forced to use !important on every declaration. This happens because our component styles (injected via rollup-plugin-styler at runtime) load after the consuming app's styles in an unpredictable order, and since both sides write plain CSS with no explicit priority rules, the last one loaded wins. There is no reliable way for consumers to win this specificity battle without !important.

Solution:

We wrap all component styles in a single named CSS @layer called motif-ui directly in the SCSS source files. Each component's selectors are placed inside @layer motif-ui { }, while SCSS compile-time declarations (@use, @forward, $variables, @mixin, @function) remain outside the layer block as required by the SCSS spec.

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Screenshots / Preview

Before:

```code
.mtf-Badge { padding: 8px !important; }
```
After:

```code
.mtf-Badge { padding: 8px }
```

## Checklist

## How to Test

Tests are completed with @sengulnecmi inside real project. There was no error, complex components are also tested with different css styles. No need for !important after all.

#EDKUI-1363


src/lib/components/Motif/Icon/MotifIcon.module.scss
```diff
@@ -11,7 +11,7 @@
 
 .motifIconsDefault {
   font-family: motif-icons-default, serif !important;
-  speak: none;
+  speak: never;
```
atakanbayrak, line 14: I would like to know that this refactor is legit ?

rollup.config.mjs
```diff
@@ -85,6 +86,7 @@ export default [
       commonjs(),
       json(),
       styles({
+        //mode: ["extract", "dist/bundle.css"], //TODO: Lets keep it like this unless there is problem.
```
atakanbayrak: I will fix this with incoming commit. I am waiting review to fix it for avoiding commit garbage.

src/lib/components/Motif/Icon/MotifIcon.module.scss
aktasmehmet: it is legit. none is replaced with never

aktasmehmet (approved):

ZehranurC (approved):

PR #67, opened 2026-05-08
fix(NavBar): search is improved

aktasmehmet:
## Description

Previously, NavBar search only depended on the "enter" key press to submit. Now, it has new features:

- Enter Key press to submit (change of previous prop name)
- Icon click to submit
- New ``results`` prop to put results inside search to show the results like an autocomplete
- Click event for the search results
- ``searching `` prop to disable the search box and show a spinner
- ``visibleContainerSize`` to set the height of the results container

Other Information:

- Responsive version is also tested.
- ``loader`` prop is introduced to Motif/InputText
- Visible type information in Storybook is updated.
- New `Search` story is added in order to test the features in detail.
- Old test are updated
- New tests are added to cover all of the features of the search.

## Type of Change

- [x] 🆕 New Component
- [x] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [x] 🧹 Code Refactoring

## Screenshots / Preview

https://github.com/user-attachments/assets/35481443-2caa-467c-9615-7f5ae73eae37

Screenshot 2026-05-07 at 16 03 44
Screenshot 2026-05-07 at 16 03 34
Screenshot 2026-05-07 at 16 03 23

## Checklist

## How to Test

You can use **Story -> NavBar/Search**

Closes #EDKUI-674: Internal purposes only


src/lib/components/Motif/InputText/InputText.module.scss
```diff
@@ -1,7 +1,7 @@
 @use "@styles" as *;
 
-// size, horizontalPadding, verticalPadding
-$sizes: (("xs", "8x", "4x"), ("sm", "12x", "6x"), ("md", "16x", "8x"), ("lg", "20x", "10x"));
+// size, horizontalPadding, verticalPadding, loaderRadius
+$sizes: (("xs", "8x", "4x", "18x"), ("sm", "12x", "6x", "20x"), ("md", "16x", "8x", "22x"), ("lg", "20x", "10x", "24x"));
```
atakanbayrak, line 4: I don't see that these sizes are reflected to Figma, did task open for this situation?

src/lib/components/Motif/InputText/InputText.module.scss
```diff
@@ -108,7 +109,23 @@ $sizes: (("xs", "8x", "4x"), ("sm", "12x", "6x"), ("md", "16x", "8x"), ("lg", "2
...
+    @keyframes spin {
+      from {
+        transform: rotate(0deg);
+      }
+      to {
+        transform: rotate(360deg);
+      }
+    }
+    animation: spin 1s linear infinite;
+    transform-origin: 50% 50%;
```
atakanbayrak: We could create mixin from this and use it project and here either.

src/lib/components/NavBar/components/NavBarSearchResults.tsx
```diff
@@ -0,0 +1,23 @@
+import styles from "@/components/NavBar/NavBar.module.scss";
```
atakanbayrak, line 1: I understand that this component is created because we didnt want to use main components together but shouldn't we find better approach instead of this solution. Because there are strict sizes as given in line 5 and component logic is similar to Dropdown, it will be hard to maintain and understand the code inside in this internal components in the future in my opinion.

src/lib/components/NavBar/NavBar.test.tsx
```diff
@@ -57,27 +58,102 @@ describe("NavBar", () => {
...
+    await userEvent.click(document.body);
+    expect(queryByText("Result 1")).not.toBeInTheDocument();
+  });
+
+  it("should close the search results dropdown when a search result is clicked", () => {
+    const { queryByText, getByText } = render(<NavBar search={{ results: [{ text: "Result 1" }] }} />);
+    expect(getByText("Result 1")).toBeInTheDocument();
+    fireEvent.click(getByText("Result 1"));
+    expect(queryByText("Result 1")).not.toBeInTheDocument();
+  });
```
atakanbayrak, line 144: We could merge these test I guess, I know that their purpose is different but logic is the same.

atakanbayrak (changes requested):

src/lib/components/NavBar/components/NavBarSearchResults.tsx
aktasmehmet: You are right. I'm gonna move that part into the scss file of the main component

src/lib/components/NavBar/NavBar.test.tsx
aktasmehmet: I did this on purpose. Merging them with "and" is a good approach the distinguish the purpose and especially the trigger of the tests

src/lib/components/Motif/InputText/InputText.module.scss
aktasmehmet: I've just taken the notes on this. Will be available next or the other sprint. Going to create the tasks this week.

Resolving this.

src/lib/components/NavBar/components/NavBarSearchResults.tsx
aktasmehmet: Naah, just checked but right now it is impossible to move it to scss side because of the fixed sizes. But you are right. In fact, this part will be the base for a future autocomplete component. I have a task in my pocket to create a common container component for dropdown, select and autocomplete...

atakanbayrak (approved):

ZehranurC (approved):
Everything looks perfect.

PR #68, opened 2026-05-11
fix(Form): validateOnChange feature

aktasmehmet:
## Description

There is a need for form component to validate the inputs when they change. This PR solves that issue.

- validateOnChange prop is introduced
- The most important part is that due to the nature of useEffect order, initial value of the inputs are validated, which is not wanted. So, there is a timeout mechanism included. This provides a macrotask, so the item is not validated initially.
- Tests are added
- There are some refactoring due to the seperation of the "whole form validate" - "item only validate"
- Documentation is updated (Including some out-of-PR form mdx updates)

## Type of Change

- [ ] 🆕 New Component
- [x] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [x] 🧹 Code Refactoring

## Screenshots / Preview

https://github.com/user-attachments/assets/c71bd7fb-25e0-40a9-a579-8dc2d1b09bd1

## Checklist

## How to Test

by setting ``validateOnChange`` to true and try to change an input.

### DIFF
--- a/src/docs/v1/form/formInputs.mdx
+++ b/src/docs/v1/form/formInputs.mdx
@@ -25,6 +25,14 @@ of a form using these components, please refer to [Components/Form](?path=/docs/
 </Form.Field>
 ```
 
+### InputDateRange [[↗]](?path=/docs/components-inputdaterange--docs)
+
+```jsx
+<Form.Field name="filterDateRange" label="Dates to Filter">
+  <InputDateRange />
+</Form.Field>
+```
+
 ### InputDateTime [[↗]](?path=/docs/components-inputdatetime--docs)
 
 ```jsx
@@ -120,11 +128,32 @@ of a form using these components, please refer to [Components/Form](?path=/docs/
 </Form.Field>
 ```
 
+### Textarea [[↗]](?path=/docs/components-textarea--docs)
+
+```jsx
+<Form.Field name="desc" label="Description">
+  <Textarea />
+</Form.Field>
+```
+
 ### UploadInput [[↗]](?path=/docs/components-upload-input--docs)
 
 ```jsx
 <Form.Field name="avatar" label="Profile Picture">
   <UploadInput
+    uploadRequest={{
+      url: "/api/upload",
+      method: "POST"
+    }}
+  />
+</Form.Field>
+```
+
+### UploadDragger [[↗]](?path=/docs/components-upload-dragger--docs)
+
+```jsx
+<Form.Field name="files" label="Additional Files">
+  <UploadDragger
     uploadRequest={{
       url: "/api/upload",
       method: "POST"
--- a/src/docs/v1/form/formInputs.mdx
+++ b/src/docs/v1/form/formInputs.mdx
@@ -141,10 +141,8 @@ of a form using these components, please refer to [Components/Form](?path=/docs/
 ```jsx
 <Form.Field name="avatar" label="Profile Picture">
   <UploadInput
-    uploadRequest={{
-      url: "/api/upload",
-      method: "POST"
-    }}
+    uploadRequest={{ url: "/api/upload", method: "POST" }}
+    deleteRequest={{ url: "/api/delete", method: "POST" }}
   />
 </Form.Field>
 ```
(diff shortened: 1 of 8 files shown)


src/docs/v1/form/formInputs.mdx
```diff
@@ -133,6 +149,19 @@ of a form using these components, please refer to [Components/Form](?path=/docs/
...
 
+### UploadDragger [[↗]](?path=/docs/components-upload-dragger--docs)
+
+```jsx
+<Form.Field name="files" label="Additional Files">
+  <UploadDragger
+    uploadRequest={{
+      url: "/api/upload",
+      method: "POST"
+    }}
```
atakanbayrak: deleteRequest is missing in here.

src/docs/v1/form/validation.mdx
```diff
@@ -217,3 +223,21 @@ You can add multiple validation rules to a field. Those may include a combinatio
   <InputText />
 </Form.Field>
 ```
+
+## Validate on Change
```
atakanbayrak, line 227: I am not quite sure that adding this prop as optional. I mean if somebody would like to add validation where is the purpose to add validations at the end of submit click? Basically, if there is validation we could automatically add this prop as default. We could discuss this with design team because for my opinion if there is validation I would like to see it when I was typing not end of the submit click.

src/docs/v1/form/validation.mdx
```diff
@@ -217,3 +223,21 @@ You can add multiple validation rules to a field. Those may include a combinatio
...
+
+By default, validations are triggered when the form is submitted. However, you can enable real-time validation by using
+the `validateOnChange` prop on the `<Form>` component. When this prop is set to `true`, validations will be executed
+whenever the value of a field changes, providing immediate feedback to users.
+
+```jsx
+<Form validateOnChange onSubmit={() => {}}>
```
atakanbayrak, line 234: For this usage, everything works normally but when I just delete validateOnChange prop, it still validates on change until I rerender page.

src/lib/components/Form/test/Form.test.tsx
```diff
@@ -1072,4 +1072,183 @@ describe("Form", () => {
...
+            <PinCode.Item />
+          </PinCode>
+        </Form.Field>
+      </Form>,
+    );
+
+    const user = userEvent.setup();
+
+    // InputText: type 2 chars → error; type 1 more → clear
+    const textInput = screen.getByTestId("inputItem").querySelector("input")!;
```
atakanbayrak: Usage of "as HTMLInputElement" instead of "!" is better approach. This comment is legit for other usage under this file.

src/lib/components/Form/Form.mdx
```diff
@@ -136,3 +101,5 @@ also a layout customization option using the `<Grid>` component inside **Form**.
 ```
 
 > **Note:** For more please refer to [Form](?path=/docs/form-form--docs)
+
+validateOnChange ekleeeeeee
```
atakanbayrak: yes we definitely should do it. 🙂

atakanbayrak (changes requested):

src/docs/v1/form/validation.mdx
aktasmehmet: Disagree on this. In fact this is actually a bad UX to show user an error message while s/he is still typing. In fact, the correct way of doing this to validate onblur. But this is what is asked to add 🤷 

So, I dont like this feature. I dont like (x2) to make this feature as default :)

src/lib/components/Form/Form.mdx
aktasmehmet: ahahhaha :)

src/lib/components/Form/test/Form.test.tsx
atakanbayrak: This one and other ones are still here but will there be new commit ?

aktasmehmet: nope, my bad, though there was only one.

aktasmehmet: just sent the new commit

atakanbayrak (approved):

ZehranurC (approved):
lgtm

PR #73, opened 2026-05-22
refactor(PinCode): custom input is replaced

aktasmehmet:
## Description

This PR is to replace the custom input text in PinCode component with the common internal InputText component.

Necessary refactors in the component and test updates are also applied.

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [x] 🧹 Code Refactoring

## Screenshots / Preview

No visual changes expected

## Checklist

## How to Test

PinCode story might be used.


src/lib/components/PinCode/__snapshots__/PinCode.test.tsx.snap
```diff
@@ -1,25 +1,31 @@
...
-    <input
-      class="input md"
-      data-testid="pinCodeItem"
-      maxlength="1"
-      type="text"
-      value=""
-    />
+    <div
+      class="Root input"
+      data-testid="inputItem"
```
atakanbayrak, line 11: I know that avoiding from using this data-testid is not possible but maybe we can just use parent element of inputItem to reach pinCode data-testid at least. So we could delete that data-testid, because there are so many data-testid inside DOM rn.

src/lib/components/PinCode/components/PinCodeItemHOC.tsx
```diff
@@ -24,16 +26,16 @@ const PinCodeItemHOC = (props: PropsWithRef<PinCodeItemHOCProps, HTMLInputElemen
   }, [letterCase, setCaseSensitiveValue, value]);
 
   const changeHandler = useCallback(
-    (e: ChangeEvent<HTMLInputElement>) => {
-      const newValue = setCaseSensitiveValue(e.target.value);
+    (val?: string) => {
```
atakanbayrak, line 29: This val prop is optional but same value prop which is located at line 16 is not optional, maybe we should consider this difference.

atakanbayrak (changes requested):

src/lib/components/PinCode/__snapshots__/PinCode.test.tsx.snap
aktasmehmet: this test id is not coming from PinCode. It's coming from InputText inside PinCode. So, there is no test id for PinCode itself and it is not removable from InputText.

src/lib/components/PinCode/components/PinCodeItemHOC.tsx
aktasmehmet: Actually I've already implemented that here:

``const newValue = setCaseSensitiveValue(val || "");```

but forgot to add to the method signature. Thx for telling me.

In fact, I've improved that a little bit and moved the optional "" into a default value in method signature.

atakanbayrak (dismissed):

src/lib/components/PinCode/components/PinCodeItemHOC.tsx
```diff
@@ -24,16 +26,16 @@ const PinCodeItemHOC = (props: PropsWithRef<PinCodeItemHOCProps, HTMLInputElemen
...
+      const newValue = setCaseSensitiveValue(val);
       newValue.length === 1 && focusNextInput(index);
     },
     [focusNextInput, index, setCaseSensitiveValue],
   );
 
   const keyUpHandler = useCallback(
     (e: KeyboardEvent<HTMLInputElement>) => {
-      e.key === "Backspace" && !value && focusPreviousInput(index, value);
+      e.key.toLocaleLowerCase() === "backspace" && !value && focusPreviousInput(index, value);
```
ZehranurC: This is not directly related to this PR, but you might consider removing the value parameter from focusPreviousInput type . The value parameter is defined in the types.ts but never used in the implementation.--> focusPreviousInput(index: number, value: string)

aktasmehmet: fixed along with some other conflict.

atakanbayrak (approved):

ZehranurC (approved):

PR #75, opened 2026-06-01
fix(Form): extra buttons and submit button removal

aktasmehmet:
## Description

This PR adds a new feature to Form component to add custom buttons to the submit button area.

Another important feature is that, from now on, ``onSubmit`` prop is optional and "not giving" onSubmit will make submit button disappear.

## Type of Change

- [ ] 🆕 New Component
- [x] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Screenshots / Preview

Screenshot 2026-06-01 at 14 29 30

## Checklist

## How to Test

- You may use new ``alternateButtons`` prop to test.
- Make ``onSubmit`` prop ``undefined`` to test "removal of the submit button"


src/lib/components/Form/types.ts
```diff
@@ -29,6 +30,12 @@ export type InputValue = string | boolean | object | [] | FileType | number | un
 export type InputSize = Size4SM;
 export type Orientation = "horizontal" | "vertical";
 export type NameInputValue = Record<string, InputValue>;
+export type AlternateFormButton = {
```
atakanbayrak: We should consider reflect the properties which were defined in Button types, because there will be mismatch after Button and Form Custom Buttons are used together. For example, Buttons will be outlined shape and this custom buttons will not have that property. So we should think the logic behind this alternate custom buttons I guess.

src/lib/components/Form/types.ts
```diff
@@ -29,6 +30,12 @@ export type InputValue = string | boolean | object | [] | FileType | number | un
...
+  label: string;
+  onClick: () => void;
+  icon?: IconGlobalType;
+  variant?: "primary" | "secondary" | "info" | "success" | "warning" | "danger";
```
atakanbayrak: My same opinion which I declared in line 33 also applies to this line because of so many variant naming. This one is different from the others because it just reflects button variants from button types, any mismatch will come back as a work to us.

src/lib/components/Form/types.ts
aktasmehmet: You are absolutely right, with one little important situation. Since Form has its own size prop, size of the buttons should not be affective.

I've made the change and let the form to decide the size. It overrides the one in <Button>

src/lib/components/Form/types.ts
aktasmehmet: Your previous comment and the solution there alredy solves this issue. So, I am resolving it.

atakanbayrak (approved):

ZehranurC (approved):
lgtm 🚀

PR #76, opened 2026-06-01
fix: library font family is added to storybook environment with higher hierarchy

atakanbayrak:
## Description

Storybook font family is inherited from its own dist file, so font-family which is used for our library is reflected inside storybook components and only components itself.

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [x] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Screenshots / Preview

font-family-2

fontfamily

## Checklist

## How to Test

Local storybook could be checked.


aktasmehmet (approved):
🚀

PR #77, opened 2026-06-02
test: localisaton tests are added based on features

atakanbayrak:
## Description

Localization tests are added with different scenarios based on our features. 

createTranslator() unit tests (14 tests)
  - Resolves the correct English and Turkish strings for keys
  - Covers all top-level sections (g, form, table, validation, upload, misc)
  - Interpolates single and multiple {{param}} placeholders
  - Keeps placeholders intact when params are missing
  - Text overrides take precedence over built-in locale strings
  - Only specified override keys change — others fall through to the locale
  - Interpolation works inside override strings
  - Unknown keys fall back to the raw key string

  MotifProvider integration tests (12 tests)
  - String locale prop sets context.locale and routes t() to the correct locale file
  - Defaults to English with no locale prop
  - Object prop { locale, texts } applies overrides, non-overridden keys stay from built-in locale
  - Object prop without locale (custom locale) falls back to English for missing keys
  - setLocale at runtime switches both locale state and translation output
  - External t prop is used instead of built-in; params are forwarded; locale/setLocale don't override it

  Locale parity tests (4 tests)
  - Every key in en.json exists in tr.json and vice versa
  - All leaf keys resolve to string values in both locales

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Checklist

## How to Test

npm run test:ci

### DIFF
--- a/src/i18n/locales/en.json
+++ b/src/i18n/locales/en.json
@@ -61,6 +61,7 @@
     "chooseOrDragImage": "Choose or drag an image"
   },
   "misc": {
-    "playgroundDescription": "This is the playground section to try your components..."
+    "playgroundDescription": "This is the playground section to try your components...",
+    "localizationTestValue": "Localization test value"
   }
 }
--- a/src/i18n/localization.test.tsx
+++ b/src/i18n/localization.test.tsx
@@ -8,6 +8,12 @@ import en from "./locales/en.json";
 import tr from "./locales/tr.json";
 import type { DeepPartial, LocaleShape } from "../lib/types";
 
+const flattenKeys = (obj: Record<string, unknown>, prefix = ""): string[] =>
+  Object.entries(obj).flatMap(([k, v]) => {
+    const full = prefix ? `${prefix}.${k}` : k;
+    return typeof v === "object" && v !== null ? flattenKeys(v as Record<string, unknown>, full) : [full];
+  });
+
 describe("Localization", () => {
   it("should return the English string for the given key", () => {
     const t = createTranslator("en");
@@ -86,7 +92,7 @@ describe("Localization", () => {
   });
 
   it("should fall back to English when a key is missing in the selected locale", () => {
-    const t = createTranslator("tr", {});
+    const t = createTranslator("en", {});
     expect(t("misc.playgroundDescription")).toBe(locales.tr.misc.playgroundDescription);
   });
 
@@ -96,20 +102,7 @@ describe("Localization", () => {
     expect(result).toBe("nonexistent.key");
   });
 
-  it("should fall back to English for missing keys when only localeTexts is supplied to custom locale", () => {
-    const t = createTranslator("en", { upload: { selectFile: "Choisir un fichier" } });
-    expect(t("upload.selectFile")).toBe("Choisir un fichier");
-    expect(t("g.save")).toBe(en.g.save);
-  });
-
   it("should expose the correct locale string via useMotifContext", () => {
-    const { result } = renderHook(() => useMotifContext(), {
-      wrapper: ({ children }: PropsWithChildren) => <MotifProvider locale="tr">{children}</MotifProvider>,
-    });
-    expect(result.current.locale).toBe("tr");
-  });
-
-  it("should translate keys using the selected locale", () => {
     const { result } = renderHook(() => useMotifContext(), {
       wrapper: ({ children }: PropsWithChildren) => <MotifProvider locale="tr">{children}</MotifProvider>,
     });
@@ -278,12 +271,6 @@ describe("Localization", () => {
     expect(result.current.t("validation.minLength", { min: 8 })).toBe("Ce champ doit comporter au moins 8 caractères");
   });
 
-  const flattenKeys = (obj: Record<string, unknown>, prefix = ""): string[] =>
-    Object.entries(obj).flatMap(([k, v]) => {
-      const full = prefix ? `${prefix}.${k}` : k;
-      return typeof v === "object" && v !== null ? flattenKeys(v as Record<string, unknown>, full) : [full];
-    });
-
   it("should have all English keys present in the Turkish locale", () => {
     const enKeys = flattenKeys(en);
     const trKeys = new Set(flattenKeys(tr));
@@ -298,18 +285,10 @@ describe("Localization", () => {
     expect(missing).toEqual([]);
   });
 
-  it("should have string values for all leaf keys in English locale", () => {
+  it("should have string values for all leaf keys in given language locale", () => {
     const enKeys = flattenKeys(en);
     enKeys.forEach(key => {
       const t = createTranslator("en");
-      expect(typeof t(key as Parameters<typeof t>[0])).toBe("string");
-    });
-  });
-
-  it("should have string values for all leaf keys in Turkish locale", () => {
-    const trKeys = flattenKeys(tr);
-    trKeys.forEach(key => {
-      const t = createTranslator("tr");
       expect(typeof t(key as Parameters<typeof t>[0])).toBe("string");
     });
   });
--- a/src/i18n/localization.test.tsx
+++ b/src/i18n/localization.test.tsx
@@ -1,0 +1,286 @@
+import "@testing-library/jest-dom";
+import { act, renderHook } from "@testing-library/react";
+import { PropsWithChildren } from "react";
+import { createTranslator } from "./translate";
+import { locales } from "./locales/index";
+import MotifProvider, { useMotifContext } from "../lib/motif/context/MotifProvider";
+import en from "./locales/en.json";
+import tr from "./locales/tr.json";
+
+describe("createTranslator()", () => {
+  describe("built-in locale resolution", () => {
+    it("should return the English string for the given key", () => {
+      const t = createTranslator("en");
+      expect(t("g.save")).toBe(en.g.save);
+    });
+
+    it("should return the Turkish string for the given key", () => {
+      const t = createTranslator("tr");
+      expect(t("g.save")).toBe(tr.g.save);
+    });
+
+    it("should return translations for all top-level sections", () => {
+      const t = createTranslator("tr");
+      expect(t("validation.required")).toBe(tr.validation.required);
+      expect(t("upload.selectFile")).toBe(tr.upload.selectFile);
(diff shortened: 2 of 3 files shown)


src/i18n/localization.test.tsx
```diff
@@ -0,0 +1,286 @@
+import "@testing-library/jest-dom";
+import { act, renderHook } from "@testing-library/react";
+import { PropsWithChildren } from "react";
+import { createTranslator } from "./translate";
+import { locales } from "./locales/index";
+import MotifProvider, { useMotifContext } from "../lib/motif/context/MotifProvider";
+import en from "./locales/en.json";
+import tr from "./locales/tr.json";
+
+describe("createTranslator()", () => {
```
atakanbayrak: There are different describes which was created by AI, I know that our pattern doesn't match with this structure but it looks well designed for me, so I keep it like that. Considerable

src/i18n/localization.test.tsx
```diff
@@ -0,0 +1,286 @@
...
+
+    it("should interpolate params in override strings", () => {
+      const t = createTranslator("en", { validation: { minLength: "Min {{min}} chars" } });
+      expect(t("validation.minLength", { min: 5 })).toBe("Min 5 chars");
+    });
+  });
+
+  describe("fallback behaviour", () => {
+    it("should fall back to English when a key is missing in the selected locale", () => {
+      // Simulate a partial locale by casting – inject a locale without 'misc'
```
atakanbayrak: Comment lines will be deleted incoming commit. Incoming commit wait for review btw.

hatesalp:
The test suite looks well organized and covers the main MotifProvider use cases nicely. createTranslator is tested well for locale resolution, parameter interpolation, custom text overrides, and fallback behavior. The MotifProvider tests also look good, especially the use of renderHook wrappers and act for state changes.

The locale parity tests are also very useful. They can help catch missing translation keys between en and tr files in the future. The tests look clean and easy to maintain 👍

src/i18n/localization.test.tsx
```diff
@@ -0,0 +1,316 @@
...
+  });
+
+  it("should override only the provided keys and keep others from the locale", () => {
+    const override = "Dosya Seçiniz";
+    const t = createTranslator("tr", { upload: { selectFile: override } });
+    expect(t("upload.selectFile")).toBe(override);
+    expect(t("g.save")).toBe(tr.g.save);
+  });
+
+  it("should override deeply nested keys", () => {
```
aktasmehmet, line 83: This test perfectly covers what it has to cover 👍 

So, It would be nicer to to use not-nested keys in :

``should override only the provided keys and keep others from the locale``
and 
``should use the override string instead of the built-in one``

to make them look simpler.

src/i18n/localization.test.tsx
```diff
@@ -0,0 +1,316 @@
...
+  });
+
+  it("should interpolate params in override strings", () => {
+    const t = createTranslator("en", { validation: { minLength: "Min {{min}} chars" } });
+    expect(t("validation.minLength", { min: 5 })).toBe("Min 5 chars");
+  });
+
+  it("should fall back to English when a key is missing in the selected locale", () => {
+    const t = createTranslator("tr", {});
+    expect(t("misc.playgroundDescription")).toBe(locales.tr.misc.playgroundDescription);
```
aktasmehmet: You are checking ``locales.tr``here. I guess it should be "en".

Besides, createTranslator doesnt need the 2nd param here I guess.

src/i18n/localization.test.tsx
```diff
@@ -0,0 +1,316 @@
...
+  });
+
+  it("should return the raw key when the key is not found in any locale", () => {
+    const t = createTranslator("en");
+    const result = t("nonexistent.key" as Parameters<typeof t>[0]);
+    expect(result).toBe("nonexistent.key");
+  });
+
+  it("should fall back to English for missing keys when only localeTexts is supplied to custom locale", () => {
```
aktasmehmet: I think this is redundant. ``should fall back to English when a key is missing in the selected locale`` should already cover what this does.

src/i18n/localization.test.tsx
```diff
@@ -0,0 +1,316 @@
...
+    const t = createTranslator("en", { upload: { selectFile: "Choisir un fichier" } });
+    expect(t("upload.selectFile")).toBe("Choisir un fichier");
+    expect(t("g.save")).toBe(en.g.save);
+  });
+
+  it("should expose the correct locale string via useMotifContext", () => {
+    const { result } = renderHook(() => useMotifContext(), {
+      wrapper: ({ children }: PropsWithChildren) => <MotifProvider locale="tr">{children}</MotifProvider>,
+    });
+    expect(result.current.locale).toBe("tr");
```
aktasmehmet: Checking at least a single text in Turkish here would be nice.

aktasmehmet: You may just remove the test ``should translate keys using the selected locale`` and check ``expect(result.current.t("g.save")).toBe(tr.g.save);`` here I guess?

src/i18n/localization.test.tsx
```diff
@@ -0,0 +1,316 @@
...
+    });
+  });
+
+  it("should have string values for all leaf keys in Turkish locale", () => {
+    const trKeys = flattenKeys(tr);
+    trKeys.forEach(key => {
+      const t = createTranslator("tr");
+      expect(typeof t(key as Parameters<typeof t>[0])).toBe("string");
+    });
+  });
```
aktasmehmet: if one of these tests fail, then the other one will also fail. Because you've already checked them in previous 2 tests. So, one of them would be redundant I guess. If you'll remove one, then please consider to change the name accordingly (not only testing english or turkish, a language let's say.)

src/i18n/localization.test.tsx
```diff
@@ -0,0 +1,316 @@
...
+    });
+
+    expect(result.current.t("validation.minLength", { min: 8 })).toBe("Ce champ doit comporter au moins 8 caractères");
+  });
+
+  const flattenKeys = (obj: Record<string, unknown>, prefix = ""): string[] =>
+    Object.entries(obj).flatMap(([k, v]) => {
+      const full = prefix ? `${prefix}.${k}` : k;
+      return typeof v === "object" && v !== null ? flattenKeys(v as Record<string, unknown>, full) : [full];
+    });
```
aktasmehmet: It would be nicer to move this out of describe block, or at least to top of it.

aktasmehmet (commented):
I didn't see any missing scenario, perfect work 💪 

There are just my standard "I've reviewed this code" kind of requests :D

src/i18n/localization.test.tsx
atakanbayrak: Other keys are also have same depth with this upload selectFile key depth value so it will change nothing to do it.

aktasmehmet: yes, I mean, the other tests' main purpose is not to test the depth. Even though, they still use a nested value to check. (It is not wrong, I just say that checking a flat value will be less complicated, that's all.) You may keep or change. It is up to you 😉 

"Resolve" button is your reponsibility now :)

src/i18n/localization.test.tsx
aktasmehmet: There is something wrong with this test. I dont understand why this passes. Maybe a false positive.

The test should be:

```
const t = createTranslator("**tr**"); //different language than the default
expect(t("misc.playgroundDescription")).toBe(**en**.misc.playgroundDescription); // fallback -> English?
```

right?

src/i18n/localization.test.tsx
atakanbayrak: roll the dice then. :D

src/i18n/localization.test.tsx
```diff
@@ -92,8 +92,8 @@ describe("Localization", () => {
   });
 
   it("should fall back to English when a key is missing in the selected locale", () => {
-    const t = createTranslator("en", {});
-    expect(t("misc.playgroundDescription")).toBe(locales.tr.misc.playgroundDescription);
+    const t = createTranslator("tr");
+    expect(t("misc.playgroundDescription")).toBe(locales.en.misc.playgroundDescription);
```
aktasmehmet: New code is fine but the issue is that, previous one should not work. But somehow, it passed and come here.

This should be investigated. When translator is English and key is not found (or found) it was checking a Turkish text and it passed. This should not be!

aktasmehmet (changes requested):

src/i18n/localization.test.tsx
atakanbayrak: I fix this problem with incoming commit. Main issue is that while we were defining tr.json file we gave the same value for misc.playgrounDescription. I fix that value but I cannot just add new value to tr.json file and make it absent in en.json because there is also test checks if these 2 file have same keys. So I found a solution to add localizationTestValue both json files and make it null one of them.

src/i18n/locales/en.json
```diff
@@ -1,66 +1,67 @@
...
-    "draggerMaxSize": "each up to {{maxSize}} in size",
-    "draggerMaxFile": "maximum {{maxFile}}",
-    "draggerCanUploadFiles": " files can be uploaded.",
-    "customValidationError": "An error occurred.",
-    "chooseOrDragImage": "Choose or drag an image"
-  },
-  "misc": {
-    "playgroundDescription": "This is the playground section to try your components..."
-  }
+{
```
aktasmehmet, line 1: This diff should not be like this. There is only one line added but it looks like the whole file has changed. You may refer to the tr.json changes. This should be like that.

src/i18n/localization.test.tsx
aktasmehmet: Clever, thx 😉

src/i18n/locales/en.json
atakanbayrak: I saw to same thing, looked up before but this 2 file is same with syntax and indent. I don't know how this en.json diff looks like this.

atakanbayrak: I copied the diff one with the local one but only change is new value that I add.

aktasmehmet (approved):

PR #79, opened 2026-06-04
fix: add colspan and rowspan support

ZehranurC:
## Description

Add colspan and rowspan configuration to table columns with proper cell rendering and pagination safety.

## Type of Change

- [ ] 🆕 New Component
- [x] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Screenshots / Preview

image
image

## Checklist

## How to Test

[#EDKUI-1410]


src/lib/components/Table/components/Body/TableBody.tsx
```diff
@@ -31,7 +34,9 @@ const TableBody = memo((props: Props) => {
           </td>
         </tr>
       ) : (
-        visibleRows.map((row, index) => <DataRow key={row.motifIndex} rowNumberStatic={index + 1} row={row} />)
+        visibleRows.map((row, index) => (
+          <DataRow key={row.motifIndex} rowNumberStatic={index + 1} row={row} spannedCellsMap={spannedCellsMap} />
```
atakanbayrak: I see that we use index + 1 each time for rowNumber, so I am wondering for example how am I going to show second row to user? For example, 

Image

Image

I didn't manage to show Santina, Isac, Antone at the table only the first row appears.

atakanbayrak: For empty row at least filled one should match with the rowSpan I guess,

Image

Image

atakanbayrak (changes requested):

src/lib/components/Table/components/Body/TableBody.tsx
aktasmehmet: ``spannedCellsMap``here depends on ``columns`` and ``visibleRows`` in line 19. They are both coming from the context. So, you retrieve the data from context and pass it to DataRow component which already has the same data.

So, it looks like spannedCellsMap here is redundant. You may move that into context itself and access it directly from DataRow; and remove everything here in TableBody?

src/lib/components/Table/types.ts
```diff
@@ -38,6 +38,8 @@ export type Column = {
   footer?: Footer;
   width?: string;
   filter?: boolean;
+  colSpan?: number | ((rowData: object) => number);
```
aktasmehmet, line 55: why need ``number``type here? If it is to apply colspan for every cell, then user can easily reduce the number of colums, so I dont see a need here?

src/lib/components/Table/types.ts
```diff
@@ -38,6 +38,8 @@ export type Column = {
...
+  rowSpan?: number | ((rowData: object) => number);
```
aktasmehmet, line 56: why need ``number``type here? If it is to apply rowspan for every cell, then user can easily reduce the number of data to half and provide that data. ?

src/lib/components/Table/components/Head/FilterColumnsRow.tsx
```diff
@@ -2,17 +2,22 @@ import styles from "../../Table.module.scss";
...
 const FilterColumnsRow = () => {
   const { columns, showFixedRowNumbers, selectable } = useContext(TableContext);
 
   return (
     <tr className={styles.trColumnFilters}>
-      {selectable && <th />}
-      {showFixedRowNumbers && <th />}
-      {columns.map((column, index) => {
+      {selectable && <th aria-hidden />}
+      {showFixedRowNumbers && <th aria-hidden />}
```
aktasmehmet: Why are those aria-hiddens added?

src/lib/components/Table/helper.ts
```diff
@@ -0,0 +1,96 @@
+import { Column, RowDetail } from "@/components/Table/types";
```
aktasmehmet: it would be better to move this file, constants.ts and sorking.ts into a folder called ``helper``

src/lib/components/Table/cellSpan.ts
```diff
@@ -0,0 +1,96 @@
...
+export const getSpannedCellsMap = (columns: Column[], rows?: RowDetail[]): SpannedCellsMap => {
+  const map = new Map<SpannedCellKey, ResolvedCellSpan | undefined>();
+  if (!rows) return map;
+
+  rows.forEach(row => {
+    columns.forEach((column, colIndex) => {
+      const cellKey: SpannedCellKey = `${row.motifIndex}-${colIndex}`;
+      if (map.has(cellKey)) return;
+
+      const remainingRowsCount = rows.filter(r => r.motifIndex >= row.motifIndex).length;
```
aktasmehmet: You are filtering the entire rows array for every single cell in the table. If you have a table with 500 rows and 10 columns, this .filter() function runs 5,000 times, performing up to 2.5 million row checks 🫨 

You should move this otside of the columns loop at least.

aktasmehmet: Another and more dangerous problem here is that since it uses original index order, this sill clamp all rowSpan to 1 after any sort!

Just check the ss

Image

here is the data to test:

```
<Table
        data={[
          { basarili: "Evet", nasil: "İyi", sonuc: "Mutlu" },
          { basarili: "Hayır", nasil: "Kötü", sonuc: "Mutsuz" },
          { basarili: "Evet", nasil: "Orta", sonuc: "Kararsız" },
          { basarili: "Belki", nasil: "Eh İşte", sonuc: "Belirsiz" },
          { basarili: "Hayır", nasil: "Çok Kötü", sonuc: "Çok Mutsuz" },
          { basarili: "Evet", nasil: "Mükemmel", sonuc: "Çok Mutlu" },
          { basarili: "Belki", nasil: "Fena Değil", sonuc: "Kararsız" },
          { basarili: "Hayır", nasil: "Berbat", sonuc: "Çok Mutsuz" },
          { basarili: "Evet", nasil: "Harika", sonuc: "Çok Mutlu" },
          { basarili: "Belki", nasil: "Orta", sonuc: "Kararsız" },
        ]}
        columns={[
          { dataKey: "basarili", title: "Başarılı", colSpan: obj => (obj.basarili === "Evet" ? 2 : 1) },
          { dataKey: "nasil", title: "Nasıl" },
          { dataKey: "sonuc", title: "Sonuç", rowSpan: obj => (obj.sonuc.includes("Mutlu") ? 2 : 1), sorting:{} },
        ]}
      />
```

src/lib/components/Table/components/Foot/ColumnFootArea.tsx
```diff
@@ -23,14 +24,15 @@ const ColumnFootArea = ({ background, customFooter }: Props) => {
   const colspanInfoList = useMemo(
     () =>
       columnsConsideringExtraCols?.reduceRight(
-        (acc, column) => {
+        (acc, column, index, arr) => {
+          const columnSpan = getColSpan(column, arr.length - index) ?? 1;
```
aktasmehmet: This will lead to a bug that footer colspan overflow with static colSpan over a non-footer column:

Image

Here is the sample code that generates the bug:

```
<Table
        data={[
          { name: "Alice", age: 28, amount: 100 },
          { name: "Bob", age: 34, amount: 200 },
          { name: "Carol", age: 29, amount: 230 },
        ]}
        columns={[
          { title: "Name", dataKey: "name", colSpan: 2 },
          { title: "Age", dataKey: "age" },
          { title: "Amount", dataKey: "amount", footer: { type: "sum" } },
        ]}
        border="cellBorders"
      />
```

src/lib/components/Table/cellSpan.ts
```diff
@@ -0,0 +1,96 @@
...
+      const { colSpan, rowSpan } = span;
+      map.set(cellKey, span);
+
+      for (let c = 1; c < colSpan; c++) {
+        map.set(`${row.motifIndex}-${colIndex + c}`, undefined);
+      }
+
+      for (let r = 1; r < rowSpan; r++) {
+        for (let c = 0; c < colSpan; c++) {
+          map.set(`${row.motifIndex + r}-${colIndex + c}`, undefined);
```
aktasmehmet: There is another bug probably because of the approach here. This will break the rowspan because it will create motifIndex gaps. You can reporocude the bug by filtering the column with "a" in the example below

```
<Table
        data={[
          { name: "Alice", age: 28, group: "A" },
          { name: "Bob", age: 34, group: "B" },
          { name: "Charlie", age: 29, group: "A" },
          { name: "Dave", age: 42, group: "B" },
        ]}
        columns={[
          { title: "Group", dataKey: "group", rowSpan: 2, filter: true },
          { title: "Name", dataKey: "name" },
          { title: "Age", dataKey: "age" },
        ]}
        border="cellBorders"
      />
```

Image

Here: 

1. getSpannedCellsMap sees row motifIndex=0 with rowSpan=2.
2. It marks key "1-0" as covered (motifIndex+1).
3. But the next VISIBLE row has motifIndex=2 — its key "2-0" is never marked!!!
4. DataRow for motifIndex=2 finds no map entry >>> renders Group cell again.

Andddd:

- HTML has rowspan="2" on row 0, AND row 1 (motifIndex=2)
- This also renders its own Group cell → overlapping cells, broken column alignment.

aktasmehmet (changes requested):
This is my review v1 :) There are some issues to fix. Then I'll check later.

This PR is so so so important because it directly manipulates the performance related codes of the component. We'll probably need more reviewers here.

src/lib/components/Table/types.ts
ZehranurC: There are two different needs:

1. **Static span (colSpan: 2)** — fixed across all rows, number type is sufficient
2. **Dynamic span (colSpan: (row) => row.merged ? 2 : 1)** — varies per row, function type is required

Modifying the data is incorrect because:
- Data integrity breaks (missing rows become null/undefined)
- Sorting/filtering breaks
- Component dataKey stops working

**example**
```tsx
// 4 rows of data, but Department column calculates rowSpan dynamically per group
const data = [
  { department: "Engineering", employee: "John" },  // rowSpan: 3
  { department: "Engineering", employee: "Jane" },  // (skip)
  { department: "Engineering", employee: "Bob" },   // (skip)
  { department: "Sales", employee: "Alice" }        // rowSpan: 1
];

// Halving the data makes no sense — grouping is a UI concern, not a data concern
rowSpan: (row) => row.firstInGroup ? groupSizes[row.department] : 1

File: src/lib/components/Table/types.ts (hunk 3 above)
Comment (ZehranurC) [reply src/lib/components/Table/types.ts:56]: The answer I provided for colSpan applies equally to rowSpan

File: src/lib/components/Table/components/Body/TableBody.tsx (hunk 1 above)
Comment (ZehranurC) [reply src/lib/components/Table/components/Body/TableBody.tsx]: It should be fixed after my fixes, but I'm not sure because I couldn't see the full data content. Can you verify?

Comment (ZehranurC) [reply src/lib/components/Table/components/Body/TableBody.tsx]: Necessary fixes have been made. You can review them.

File: src/lib/components/Table/helper.ts (hunk 5 above)
Comment (ZehranurC) [reply src/lib/components/Table/helper.ts]: Necessary fixes have been made. You can review them.

File: src/lib/components/Table/cellSpan.ts (hunk 6 above)
Comment (ZehranurC) [reply src/lib/components/Table/cellSpan.ts]: Necessary fixes have been made. You can review them.

Comment (ZehranurC) [reply src/lib/components/Table/cellSpan.ts]: It should be fixed after my fixes. You can review them.

File: src/lib/components/Table/components/Foot/ColumnFootArea.tsx (hunk 7 above)
Comment (ZehranurC) [reply src/lib/components/Table/components/Foot/ColumnFootArea.tsx]: It should be fixed after my fixes. You can review them.

File: src/lib/components/Table/cellSpan.ts (hunk 8 above)
Comment (ZehranurC) [reply src/lib/components/Table/cellSpan.ts]: It should be fixed after my fixes. You can review them.

File: src/lib/components/Table/helper.ts [hunk 9]
```diff
@@ -0,0 +1,130 @@
...
+};
+
+export type SpannedCellKey = `${number}-${number}`;
+export type SpannedCellsMap = Map<SpannedCellKey, ResolvedCellSpan | undefined>;
+
+export type RenderableColumn = {
+  column: Column;
+  index: number;
+  colSpan?: number;
+};
```
aktasmehmet: There is already a types.ts file, so it would be better to move all these types there and access types from a single file for easy management.

src/lib/components/Table/components/Head/FilterColumnsRow.tsx
aktasmehmet: You've removed I guess. But the next time, it would be better to write down a tiny explanation about the use and the removal 😉

src/lib/components/Table/types.ts
aktasmehmet: I dont find useful information in this answer. The example shows the usage of the function, not the number. 

I wonder that, what would be the result for the code below:

```
<Table
        data={[
          { name: "Alice", age: 28, group: "A" },
          { name: "Bob", age: 34, group: "B" },
          { name: "Charlie", age: 29, group: "A" },
          { name: "Dave", age: 42, group: "B" },
        ]}
        columns={[
          { title: "Group", dataKey: "group", rowSpan: 2 },
          { title: "Name", dataKey: "name" },
          { title: "Age", dataKey: "age" },
        ]}
        border="cellBorders"
      />
```

Afaik, this would render only Alice and Charline lines, right? So this will show 1st row and 3rd row and so on... So, what I mean is that, user can filter the data himself without using rowspan? This will not effect the filtering also?

Am I missing something?

src/lib/components/Table/types.ts
aktasmehmet: Resolving this. Same thing in the rowspan. When or if fixed, you can continue from that thread.

src/lib/components/Table/components/Foot/ColumnFootArea.tsx
```diff
@@ -10,7 +11,7 @@ type Props = {
 };
 
 const ColumnFootArea = ({ background, customFooter }: Props) => {
-  const { originalRows, columns, showFixedRowNumbers, selectable, numberOfVisibleColumns } = useContext(TableContext);
+  const { usableRows, columns, showFixedRowNumbers, selectable, numberOfVisibleColumns } = useContext(TableContext);
```
aktasmehmet: switching from ``originalRows``to ``usableRows`` causes the avg footer to be shown as **NaN** when all rows are filtered out. You may filter with "x" Please check the ss.

Here is the sample test:

```
<Table
        data={[
          { name: "Alice", score: 90 },
          { name: "Bob", score: 80 },
          { name: "Carol", score: 70 },
        ]}
        columns={[
          { title: "Name", dataKey: "name", filter: true },
          { title: "Score", dataKey: "score", footer: { type: "avg" } },
        ]}
        border="cellBorders"
      />
```

Image

aktasmehmet: Now saw that this also breaks the filtered total. Previously, it was showin the total of the whole table but now, it only shows the filtered ones as total. This is not expected. 

Here is the code (Sum should stay as 600):

```
<Table
        data={[
          { name: "Alice", amount: 100 },
          { name: "Bob", amount: 200 },
          { name: "Carol", amount: 300 },
        ]}
        columns={[
          { title: "Name", dataKey: "name", filter: true },
          { title: "Amount", dataKey: "amount", footer: { type: "sum" } },
        ]}
        border="cellBorders"
      />
```

Image

src/lib/components/Table/components/Foot/ColumnFootArea.tsx
```diff
@@ -20,52 +21,37 @@ const ColumnFootArea = ({ background, customFooter }: Props) => {
...
 
-    const { colSpan, title } = colspanInfoList[index];
-    return (
-      <th key={"foot_td" + index} {...(colSpan && { colSpan })}>
-        {column.footer ? (column.footer.render?.(getFooterValue(column)) ?? getFooterValue(column)) : colSpan > 0 && title ? title : ""}
-      </th>
-    );
-  });
+        return (
+          <th key={"foot_" + column.dataKey} {...(actualColSpan > 1 && { colSpan: actualColSpan })}>
```
aktasmehmet: This change will lead to duplicate key error. Here is the scenario:

When :

selectable=true 
AND
showFixedRowNumbers=true 
AND 
any data column has footer defined

--> columnsConsideringExtraCols prepends {title:''} and {title:'#'}, both with dataKey=undefined. key={'foot_' + column.dataKey} evaluates to 'foot_undefined' for both.

So this will cause the error:

```
<Table
        data={[
          { name: "Alice", amount: 100 },
          { name: "Bob", amount: 200 },
          { name: "Carol", amount: 300 },
        ]}
        columns={[
          { title: "Name", dataKey: "name" },
          { title: "Amount", dataKey: "amount", footer: { type: "sum" } },
        ]}
        selectable
        showFixedRowNumbers
        border="cellBorders"
      />
```

Image

aktasmehmet (changes requested):

src/lib/components/Table/helper.ts
ZehranurC: This has been fixed in the latest commit. You can verify it.

src/lib/components/Table/components/Foot/ColumnFootArea.tsx
ZehranurC: This should be fixed in the latest commit. You can verify it.

ZehranurC: This should be fixed in the latest commit. You can verify it.

src/lib/components/Table/components/Foot/ColumnFootArea.tsx
ZehranurC: This has been fixed in the latest commit. You can verify it.

src/lib/components/Table/types.ts
ZehranurC: I might be misunderstood.
rowSpan doesn't filter. it only determines how cells are displayed.

If you set **rowSpan: 2**, the Group cell spans 2 rows. But which rows appear depends on your data.


// Pass this data:
```
data={[
  { name: "Alice", age: 28, group: "A" },
  { name: "Charlie", age: 29, group: "A" },  // 2 people = rowSpan: 2 works fine
  { name: "Bob", age: 34, group: "B" },
  { name: "Dave", age: 42, group: "B" },
]}

```
// Result: Group A cell spans 2 rows ✓
If Group A has 3 people, rowSpan: 2 won't be enough. Then use dynamic rowSpan:

```
rowSpan: (row) => row.firstInGroup ? groupSize[row.group] : 1
```
It has nothing to do with filtering — rowSpan just decides how to merge cells in the rows that are already shown.

aktasmehmet: no, my misunderstanding. I was thinking that rowSpan here with just a number type will merge the whole row, not the cell. It is perfect like this. Sorry for the inconvenience.

src/lib/components/Table/components/Foot/ColumnFootArea.tsx
aktasmehmet: While testing, I've found another bug here. We are using originalRows as the condition but it this case ``[]`` is truthy and this willl cause a ``0/0 = NaN``. You might check the attached screenshot.

So, we must replace ``return originalRows`` like this:

```suggestion
        return originalRows?.length
```

Image

src/lib/components/Table/components/Foot/ColumnFootArea.tsx
```diff
@@ -54,18 +37,21 @@ const ColumnFootArea = ({ background, customFooter }: Props) => {
...
 
-    const { colSpan, title } = colspanInfoList[index];
-    return (
-      <th key={"foot_td" + index} {...(colSpan && { colSpan })}>
-        {column.footer ? (column.footer.render?.(getFooterValue(column)) ?? getFooterValue(column)) : colSpan > 0 && title ? title : ""}
-      </th>
-    );
-  });
+        return (
+          <th key={"foot_" + index} {...(actualColSpan > 1 && { colSpan: actualColSpan })}>
```
aktasmehmet: While other sub components use shared helper, here you dont use. Using the sanem function will keep the consistency:


```suggestion
          <th key={"foot_" + index} {...getSpanProps(actualColSpan)}>
```

src/lib/components/Table/components/Foot/ColumnFootArea.tsx
ZehranurC: It's really great that you noticed it . I've applied the update. It will come with the next commit.

src/lib/components/Table/components/Foot/ColumnFootArea.tsx
ZehranurC: Thank you for the suggestion. It will come with the next commit.

aktasmehmet (changes requested):
I was going to approve but we've both miss some issues. Please check Chromatic. Here is an example:

Image

aktasmehmet (changes requested):
Image

Some resolved but some is still there. It is because of the border-collapse you've put.

PS. For the future, please pay extra attention that the issues are validated before requesting a re-review 🙏

ZehranurC:
> Image
> Some resolved but some is still there. It is because of the border-collapse you've put.
> 
> PS. For the future, please pay extra attention that the issues are validated before requesting a re-review 🙏

**Footer title and value were being displayed in the same cell. Fixed by re-introducing 
the footer cell separation logic that was previously removed.**

aktasmehmet (changes requested):
I've found some issues and stated below

src/lib/components/Table/components/TableContainer.tsx
```diff
@@ -49,11 +52,12 @@ const TableContainer = (props: PropsWithRef<Props, HTMLDivElement>) => {
     distributeColsEvenly && "distributeColsEvenly",
     fluid && "fluid",
   ]);
+  const hasCellSpan = columns.some(column => column.colSpan !== undefined || column.rowSpan !== undefined);
 
   return (
     <div ref={ref} className={classNames} style={style}>
       <TableTitleSection title={title} subtitle={subtitle} />
-      <table>
+      <table className={hasCellSpan ? styles.hasCellSpan : undefined}>
```
aktasmehmet: This woud lead a table with class name "undefined". This is the whole need that we created a sanitize method.

Anyway, there is also a better approach instead of using a class name and hasCellSpan calculation here. You can use pure css with has selector to do the same thing here:

```
&.cellBorders {
    table:has([colspan], [rowspan]) {
      border-collapse: collapse;
    }
    ...
  }
  &.rowBorders {
    table:has([colspan], [rowspan]) {
      border-collapse: collapse;
    }
    ...
  }
```

aktasmehmet (changes requested):

src/lib/components/Table/components/Foot/ColumnFootArea.tsx
```diff
@@ -42,34 +44,39 @@ const ColumnFootArea = ({ background, customFooter }: Props) => {
...
       default:
         return undefined;
     }
   };
 
-  const footerCells = columnsConsideringExtraCols?.map((column, index) => {
-    if (!colspanInfoList?.[index] || (!column.footer && !colspanInfoList[index].colSpan)) {
-      return null;
-    }
+  const footerCells = columnsConsideringExtraCols
```
aktasmehmet, line 70: I feel a little bir confused about the new footer calculations. Found a bug instantly.

With the code below, footer becomes wider than the original body:

```
<Table
        data={[
          { name: "Alice Smith", age: 28, amount: 100 },
          { name: "Bob Jones", age: 34, amount: 200 },
        ]}
        columns={[
          { title: "Full Name", dataKey: "name", colSpan: 2 },
          { title: "Age", dataKey: "age" },
          { title: "Amount", dataKey: "amount", footer: { type: "sum" } },
        ]}
        border="cellBorders"
      />
```

Image

aktasmehmet (changes requested):

src/lib/components/Table/components/Foot/ColumnFootArea.tsx
ZehranurC: Refactored footer rendering from colspanInfoList to buildFooterCells helper. Pending colspan tracking ensures non-footer columns' colspan is handled correctly, fixing breaking change.

aktasmehmet: Soory but this time there is another issue there :(

Screenshot 2026-06-24 at 17 14 18

```
<Table
        data={[
          { name: "Alice", amount: 100, status: "Active" },
          { name: "Bob", amount: 200, status: "Inactive" },
        ]}
        columns={[
          { title: "Name", dataKey: "name" },
          { title: "Amount", dataKey: "amount", footer: { type: "sum" } },
          { title: "Status", dataKey: "status" },
        ]}
        border="cellBorders"
      />
```

aktasmehmet: Along with other issues, I've detected couple of more, especially on the footer area. They are edge cases but still issues.

So, I've fixed them all, added missing tests and updated the documentation. Resolving this. Can you please review my last commit?

aktasmehmet (dismissed):

atakanbayrak (dismissed):

src/lib/components/Table/components/Foot/ColumnFootArea.tsx
```diff
@@ -20,56 +21,78 @@ const ColumnFootArea = ({ background, customFooter }: Props) => {
...
       default:
         return undefined;
     }
   };
 
-  const footerCells = columnsConsideringExtraCols?.map((column, index) => {
-    if (!colspanInfoList?.[index] || (!column.footer && !colspanInfoList[index].colSpan)) {
-      return null;
-    }
+  const buildFooterCells = (renderableColumns: ReturnType<typeof getRenderableHeaderColumns>, allColumns: Column[]) => {
```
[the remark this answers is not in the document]
ZehranurC: Moved the getFooterValue function to the helper module. buildFooterCells remains in the component since it renders JSX.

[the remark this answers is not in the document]
ZehranurC: Moved the getFooterValue function to the helper module. buildFooterCells remains in the component since it renders JSX.

src/lib/components/Table/Table.module.scss
```diff
@@ -203,7 +206,7 @@ $row-color-variants: "primary", "secondary", "light", "success", "danger", "warn
   }
 
   &.rowBorders {
-    table {
+    table:has(thead [colspan], thead [rowspan], tbody [colspan], tbody [rowspan]) {
```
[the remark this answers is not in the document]
aktasmehmet: I verify this. Not double borders but no borders at all when ``border="rowBorders"``

src/lib/components/Table/components/Foot/ColumnFootArea.tsx
```diff
@@ -20,52 +20,56 @@ const ColumnFootArea = ({ background, customFooter }: Props) => {
...
+          return {
+            cells: [
+              ...acc.cells,
+              ...(acc.pending
+                ? [
+                    <th key={`foot_label_${index}`} {...getSpanProps(acc.pending)}>
+                      {column.footer.title}
+                    </th>,
+                  ]
+                : []),
```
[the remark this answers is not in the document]
aktasmehmet: This is dismissible right now.

aktasmehmet (approved):

atakanbayrak (approved):

PR #80, opened 2026-06-04
fix(Stepper): data management between steps

aktasmehmet:
## Description

Some improvement are made to make it easier to manage data and switch between steps. Here are the complete list:

- A new context approach is applied to easily manage data between steps.
- Next and Previous buttons are added
- Step switches using the top navigation has changed. Now, it is only managed by the buttons to go forward.
- Disabled steps are also taken into consideration.
- useStepper hook is intoduced for custom step flow management.
- Steps are centered (fix)
- Vertical orientation layout issues are fixed
- Related Figma task is created. 

**Step Management:**

- Stepper can still be used as a standalone component. It handles the navigation within itself.
- If a custom navigation or data management is required, then ``useStepper`` hook must be used. Return of this hook should be given as ``state`` prop to the Stepper itself to connect the hook and the Stepper. 

**Out of Context:**
- Small icon fix is applied for InputDate and InputTime

## Type of Change

- [ ] 🆕 New Component
- [x] ✨ Feature / Enhancement
- [x] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [x] 🧹 Code Refactoring

## Screenshots / Preview

Screenshot 2026-06-04 at 09 32 51

## Checklist

## How to Test

A file is attached here to test some of the scenarios. Just run the dev server and test. 

[page.tsx](https://github.com/user-attachments/files/28585022/page.tsx)

### DIFF
--- a/src/i18n/locales/tr.json
+++ b/src/i18n/locales/tr.json
@@ -11,6 +11,9 @@
     "upload": "Yükle",
     "selectPlease": "Lütfen seçiniz",
     "noData": "Veri bulunamadı...",
+    "previous": "Önceki",
+    "next": "Sonraki",
+    "finish": "Tamamla",
     "x_message": "{{count}} mesaj"
   },
   "form": {
--- a/src/lib/components/Stepper/Stepper.module.scss
+++ b/src/lib/components/Stepper/Stepper.module.scss
@@ -15,6 +15,13 @@ $types: (("dot", var(--base-sizing-12x)), ("number", var(--base-sizing-24x)), ("
 
 .stepList {
   display: flex;
+  justify-content: center;
+}
+
+.stepNavigation {
+  display: flex;
+  justify-content: space-between;
+  padding-top: var(--base-sizing-48x);
 }
 
 .stepItem {
@@ -23,6 +30,7 @@ $types: (("dot", var(--base-sizing-12x)), ("number", var(--base-sizing-24x)), ("
   .stepHeader {
     display: flex;
     align-items: center;
+    text-align: center;
     &.item-vertical {
       flex-direction: column;
       gap: var(--base-sizing-4x);
@@ -96,13 +104,24 @@ $types: (("dot", var(--base-sizing-12x)), ("number", var(--base-sizing-24x)), ("
   }
 }
 
+.stepContent {
+  display: flex;
+  flex-direction: column;
+}
+
 .vertical {
   display: grid;
+  grid-template-columns: max-content 1fr;
   row-gap: var(--base-sizing-8x);
 
   .stepList {
     grid-row: 2;
     flex-direction: column;
+  }
+
+  .stepNavigation {
+    align-self: flex-end;
+    gap: var(--base-sizing-8x);
   }
 
   .stepItem {
@@ -126,6 +145,7 @@ $types: (("dot", var(--base-sizing-12x)), ("number", var(--base-sizing-24x)), ("
 
   .stepContent {
     grid-row: 2;
+    justify-content: space-between;
   }
 }
 
--- a/src/lib/components/Stepper/Stepper.tsx
+++ b/src/lib/components/Stepper/Stepper.tsx
@@ -14,8 +14,8 @@ import { StepperContext } from "./StepperContext";
 const Stepper = (props: PropsWithRefAndChildren<StepperProps, HTMLDivElement>) => {
   const {
     children,
-    activeStep: controlledStep,
     defaultActiveStep = 0,
+    hideNavigation,
     showCount,
     variant = "primary",
     stepType = "number",
@@ -45,9 +45,8 @@ const Stepper = (props: PropsWithRefAndChildren<StepperProps, HTMLDivElement>) =
   const stepItems = Children.toArray(children) as ReactElement<StepperItemProps>[];
   const count = stepItems.length;
   const disabledSteps = stepItems.map(item => !!item.props.disabled);
-  const isControlled = controlledStep !== undefined;
 
-  const rawActiveStep = externalState ? externalState.activeStep : isControlled ? controlledStep : internalStep;
+  const rawActiveStep = externalState ? externalState.activeStep : internalStep;
   const resolvedStep = disabledSteps.findIndex((d, i) => i > rawActiveStep && !d);
   const activeStep = externalState && disabledSteps[rawActiveStep] && resolvedStep !== -1 ? resolvedStep : rawActiveStep;
 
@@ -55,15 +54,11 @@ const Stepper = (props: PropsWithRefAndChildren<StepperProps, HTMLDivElement>) =
     (index: number) => {
       if (!disabledSteps[index]) {
         const clamped = Math.max(0, Math.min(index, count - 1));
-        if (externalState) {
-          externalState.goToStep(clamped);
-        } else if (!isControlled) {
-          setInternalStep(clamped);
-        }
+        externalState ? externalState.goToStep(clamped) : setInternalStep(clamped);
         onStepChange?.(clamped);
       }
     },
-    [disabledSteps, count, externalState, isControlled, onStepChange],
+    [disabledSteps, count, externalState, onStepChange],
   );
 
   const goToNextStep = useCallback(() => {
@@ -111,12 +106,14 @@ const Stepper = (props: PropsWithRefAndChildren<StepperProps, HTMLDivElement>) =
         {activeContent && (
           <div className={styles.stepContent}>
             <div>{activeContent}</div>
-            <StepperNavigation
-              onFinishClick={onFinishClick}
-              finishButtonLabel={finishButtonLabel}
-              onNextClick={onNextClick}
-              onPrevClick={onPrevClick}
-            />
+            {!hideNavigation && (
+              <StepperNavigation
+                onFinishClick={onFinishClick}
+                finishButtonLabel={finishButtonLabel}
+                onNextClick={onNextClick}
+                onPrevClick={onPrevClick}
+              />
+            )}
           </div>
         )}
       </div>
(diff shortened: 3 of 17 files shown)


src/lib/components/Stepper/Stepper.tsx
```diff
@@ -1,43 +1,124 @@
 "use client";
```
atakanbayrak, line 1: When I use activeStep value 1 for given page.tsx. Rerender problem appears.

src/lib/components/Stepper/types.ts
```diff
@@ -36,4 +43,26 @@ export type StepperContextType = {
...
+  goToNextStep: () => void;
+  goToPrevStep: () => void;
+  disabledSteps: boolean[];
+  stepData: Partial<Record<number, Record<string, unknown>>>;
+  setStepData: (index: number, data: Record<string, unknown>) => void;
+};
+
+export type StepperNavigationProps = {
+  onFinishClick?: () => void;
+  finishButtonLabel?: string;
```
atakanbayrak, line 56: Are we sure that adding this prop because TimePicker also has 2 buttons but we don't give access to change button naming ? In my opinion, if we give this feature to user next step will be changing this buton type and shape. It will be hard to maintain and supply needing for this button.

src/lib/components/Stepper/components/StepperNavigation.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+  const isFirstStep = disabledSteps.slice(0, activeStep).every(Boolean);
+  const isLastStep = disabledSteps.slice(activeStep + 1).every(Boolean);
+
+  const handleNext = isLastStep ? onFinishClick : (onNextClick ?? goToNextStep);
+  const handlePrev = onPrevClick ?? goToPrevStep;
+
+  return (
+    <div className={styles.stepNavigation}>
+      <Button label={t("g.previous")} shape="outline" onClick={handlePrev} disabled={isFirstStep} />
```
atakanbayrak, line 20: If we could consider adding locale prop to Stepper component in the future, there will be a problem to fix this motif context based locale definition button naming because it only depends on layout and library locale rn. While rest of the component for example Turkish, these buttons will remain as library default and it will create language differentiation. So my suggestion is reverting these feature for here rn. When we define how to manage these texts, we could add it again, but for incoming bug; deleting is better option.

src/i18n/locales/tr.json
atakanbayrak, line 7: Does finish also reflects submit in here? We could consider using `submit` instead of adding new one.

src/lib/components/Stepper/StepperContext.tsx
```diff
@@ -1,10 +1,6 @@
...
 import { StepperContextType } from "./types";
 
-export const StepperContext = createContext<StepperContextType>({
-  activeStep: 0,
-  count: 0,
-  variant: "primary",
-  stepType: "number",
-  itemOrientation: "vertical",
-});
+export const StepperContext = createContext<StepperContextType | null>(null);
```
atakanbayrak, line 6: Doesn't we create stepperContextDefaultValues and use it inside this StepperContext as we already did it before at Date components? It will also give flexibility for deleting ! usage inside subcomponents.

atakanbayrak (changes requested):

src/lib/components/Stepper/Stepper.module.scss
```diff
@@ -15,6 +15,13 @@ $types: (("dot", var(--base-sizing-12x)), ("number", var(--base-sizing-24x)), ("
 
 .stepList {
```
atakanbayrak, line 16: I also would like to add that disabled step looks same with not active step. What I mean that, for example 3 is disabled and I am at step 1, step 2 and step 3 look same at the Stepper. I don't understand that 3 is disabled and 2 is accessible, so it kinda confused me while using component, I tried to reach 3 without any doubt.

src/lib/components/Stepper/types.ts
```diff
@@ -22,11 +22,18 @@ export type StepperDefaultableProps = {
   orientation?: StepperOrientation;
   itemOrientation?: StepperOrientation;
   showCount?: boolean;
+  finishButtonLabel?: string;
 };
 
 export type StepperProps = {
   activeStep?: number;
```
atakanbayrak: When I tried to access any activeStep with this prop I didn't manage to change active step into component. It looks like activeStep prop does not working at all.

atakanbayrak: When I use showCount and activeStep together, showCount value show that step number without controlling that if this step exists or not.

Image

```code
     <Stepper
        activeStep={6}
        finishButtonLabel="atakan"
        showCount
        defaultActiveStep={2}
        onNextClick={onNextClick}
        style={{ border: "solid 1px red", padding: 20, width: "100%" }}
      >
```

src/lib/components/Stepper/hooks/useStepper.ts
```diff
@@ -0,0 +1,21 @@
+"use client";
```
atakanbayrak, line 1: Standalone Stepper example inside Storybook does not work at all rn. I also see that, there are buttons that were not exist before, are we sure that adding these buttons built-in to Stepper?

atakanbayrak: Some users use it only for showing which step user are, so buttons are not must I guess.

src/i18n/locales/tr.json
aktasmehmet: Nope, it doesnt reflect submit. That is job of a form. This is the finishing of the wizard lets say.

src/lib/components/Stepper/StepperContext.tsx
aktasmehmet: This context is not exposed outside. Deleting will not cause any harm because it lives in each different stepper.

Besides, this is a very common pattern to overcome setting a useless initial value. Null is there for the initialization, thats's all.

src/lib/components/Stepper/types.ts
aktasmehmet: Type and shape can be shance via css but label is not. So, this is a must and requested. If there is a need for other components, we might implement also.

Changing labels like "Cancel" or "Send" will %99 percent global so they can be done via translation files but in this scenario, even we-developers called that button with 3 different names :)

So, this must stay as it is

src/lib/components/Stepper/Stepper.module.scss
aktasmehmet: Yes I know. This is a design decision. There is no difference in Figma as well.

Disabled step is an internal state actually. This is used to prevent a step is accessed. Not showing user that is inaccessible. User doesnt need to know that.

src/lib/components/Stepper/hooks/useStepper.ts
aktasmehmet: Storybook is fixed. 

I am considering to make those buttons optional

src/lib/components/Stepper/components/StepperNavigation.tsx
aktasmehmet: How is this possible:

"While rest of the component for example Turkish, these buttons will remain as library default"

There is no need for a component spesific locale here. I guess you are working too much on the pickers nowadays 😄

src/lib/components/Stepper/types.ts
aktasmehmet: Actually activeStep is to use the component uncontrolled. I understand the confusion.

I mean, if activeStep is set, then it basically means that "Do not get involved with my navigation, I'll handle it myself"... So, the next and previous buttons will not work because you have to use useStepper with the callbacks to manage the navigation yourself 🤷 

I still think to make this less confusing...

ZehranurC (commented):
In a scenario where all steps are disabled, shouldnt the 'Submit' button also be disabled? Or is this an intentional design decision?

Image

aktasmehmet:
> In a scenario where all steps are disabled, shouldnt the 'Submit' button also be disabled? Or is this an intentional design decision?
> 
> Image

With all due respect, what is the point of using Stepper if all of the steps are disabled? :) Anyway, keeping that button enabled all the time is a decision as you said. I'll also add an option to remove that buttons completely

src/lib/components/Stepper/types.ts
aktasmehmet: I've removed that. It is now only managed via the hook. Changes are coming...

src/lib/components/Stepper/Stepper.tsx
aktasmehmet: this will be outdated with the next commit

ZehranurC:
> > In a scenario where all steps are disabled, shouldnt the 'Submit' button also be disabled? Or is this an intentional design decision?
> > Image
> 
> With all due respect, what is the point of using Stepper if all of the steps are disabled? :) Anyway, keeping that button enabled all the time is a decision as you said. I'll also add an option to remove that buttons completely

I was thinking—maybe all scenarios need to wait for a payment or some time period to activate. But honestly, that's probably too complex of a use case. They likely wouldn't even show the Stepper in that situation.

ZehranurC (commented):
I've tested almost all the scenarios and everything looks correct and working as expected. I think you will push one more commit with any remaining fixes, and then I'll approve this PR. Great work!

src/lib/components/Stepper/types.ts
aktasmehmet: Can you open a task for this one to make the UI team create a design for that? I tried to fix but since the last step width is not fixed, it does not look aligned. We need a design solution to this. 

Please dont forget to attach screenshot and a good description 😉

ZehranurC (approved):
lgtm 🚀

atakanbayrak (approved):

PR #81, opened 2026-06-08
fix: localization support for date components and refactoring for tests

atakanbayrak:
## Description

Removed component-level default localization values. Components now use the localization provided by the underlying library when no locale prop is passed, while still allowing consumers to override localization by providing a custom locale prop.

Changes
- Removed default localization assignments from components.
- Updated locale handling to defer to the library's default localization.
- Preserved support for custom localization via the locale prop.
- Test cases are fixed without using constant localisation values.
- Context default values are fixed without usage promised localisation value.
- Clear and OK buttons are refactored to convert library localisation values.
- getDateLocal() function is created for calling local value without parsing every time inside component level and also using inside test cases to call local value each time.
- en.json and tr.json file is refactored to store values for date components. 
- localisation supports string[] with incoming changes, previous version was only supporting string values.
- Storybook stories are checked with the correct localisation values.

## Type of Change

- [ ] 🆕 New Component
- [x] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Screenshots / Preview

Screenshot 2026-06-08 at 09 41 06
Screenshot 2026-06-08 at 09 40 52
Screenshot 2026-06-08 at 09 40 43
Screenshot 2026-06-08 at 09 40 34

## Checklist

## How to Test

```code
<Grid>
        <Grid.Col>
          <Grid.Row>
            <DatePicker
              locale={{
                months: ["Ocak", "Şubat", "Mart", "Nisan", "Mayıs", "Haziran", "Temmuz", "Ağustos", "Eylül", "Ekim", "Kasım", "Aralık"],
                monthsShort: ["Oca", "Şub", "Mar", "Nis", "May", "Haz", "Tem", "Ağu", "Eyl", "Eki", "Kas", "Ara"],
                weekDays: ["Pz", "Pt", "Sa", "Çr", "Pe", "Cu", "Ct"],
                firstDayOfWeek: 1,
              }}
            />
            <DatePicker />
          </Grid.Row>
          <Grid.Row>
            <DateRangePicker
              locale={{
                months: ["Ocak", "Şubat", "Mart", "Nisan", "Mayıs", "Haziran", "Temmuz", "Ağustos", "Eylül", "Ekim", "Kasım", "Aralık"],
                monthsShort: ["Oca", "Şub", "Mar", "Nis", "May", "Haz", "Tem", "Ağu", "Eyl", "Eki", "Kas", "Ara"],
                weekDays: ["Pz", "Pt", "Sa", "Çr", "Pe", "Cu", "Ct"],
                firstDayOfWeek: 1,
                today: "Bugün",
                last: "Son",
                days: "Günler",
                choose: "Seç",
              }}
            />
            <DateRangePicker />
          </Grid.Row>
          <Grid.Row>
            <DateTimePicker />
            <DateTimePicker
              locale={{
                months: ["Ocak", "Şubat", "Mart", "Nisan", "Mayıs", "Haziran", "Temmuz", "Ağustos", "Eylül", "Ekim", "Kasım", "Aralık"],
                monthsShort: ["Oca", "Şub", "Mar", "Nis", "May", "Haz", "Tem", "Ağu", "Eyl", "Eki", "Kas", "Ara"],
                weekDays: ["Pz", "Pt", "Sa", "Çr", "Pe", "Cu", "Ct"],
                firstDayOfWeek: 1,
                hoursAbbr: "Atakan Saat",
                minutesAbbr: "Atakan Dakika",
                secondsAbbr: "Atakan saniye",
                am: "ÖÖ",
                pm: "ÖS",
              }}
            />
          </Grid.Row>
          <Grid.Row>
            <InputDate />
            <InputDate
              locale={{
                months: ["Ocak", "Şubat", "Mart", "Nisan", "Mayıs", "Haziran", "Temmuz", "Ağustos", "Eylül", "Ekim", "Kasım", "Aralık"],
                monthsShort: ["Oca", "Şub", "Mar", "Nis", "May", "Haz", "Tem", "Ağu", "Eyl", "Eki", "Kas", "Ara"],
                weekDays: ["Pz", "Pt", "Sa", "Çr", "Pe", "Cu", "Ct"],
                firstDayOfWeek: 1,
              }}
            />
          </Grid.Row>
          <Grid.Row>
            <InputDateRange />
            <InputDateRange
              locale={{
                months: ["Ocak", "Şubat", "Mart", "Nisan", "Mayıs", "Haziran", "Temmuz", "Ağustos", "Eylül", "Ekim", "Kasım", "Aralık"],
                monthsShort: ["Oca", "Şub", "Mar", "Nis", "May", "Haz", "Tem", "Ağu", "Eyl", "Eki", "Kas", "Ara"],
                weekDays: ["Pz", "Pt", "Sa", "Çr", "Pe", "Cu", "Ct"],
                firstDayOfWeek: 1,
                today: "Bugün",
                last: "Son",
                days: "Günler",
                choose: "Seç",
              }}
            />
          </Grid.Row>
          <Grid.Row>
            <InputDateTime />
            <InputDateTime
              locale={{
                months: ["Ocak", "Şubat", "Mart", "Nisan", "Mayıs", "Haziran", "Temmuz", "Ağustos", "Eylül", "Ekim", "Kasım", "Aralık"],
                monthsShort: ["Oca", "Şub", "Mar", "Nis", "May", "Haz", "Tem", "Ağu", "Eyl", "Eki", "Kas", "Ara"],
                weekDays: ["Pz", "Pt", "Sa", "Çr", "Pe", "Cu", "Ct"],
                firstDayOfWeek: 1,
                hoursAbbr: "Atakan Saat",
                minutesAbbr: "Atakan Dakika",
                secondsAbbr: "Atakan saniye",
                am: "ÖÖ",
                pm: "ÖS",
              }}
            />
          </Grid.Row>
          <Grid.Row>
            <InputTime />
            <InputTime
              locale={{
                hoursAbbr: "Atakan Saat",
                minutesAbbr: "Atakan Dakika",
                secondsAbbr: "Atakan saniye",
                am: "ÖÖ",
                pm: "ÖS",
              }}
            />
          </Grid.Row>
          <Grid.Row>
            <TimePicker />
            <TimePicker
              secondsEnabled
              locale={{
                hoursAbbr: "Atakan Saat",
                minutesAbbr: "Atakan Dakika",
                secondsAbbr: "Atakan saniye",
                am: "ÖÖ",
                pm: "ÖS",
              }}
            />
          </Grid.Row>
        </Grid.Col>
      </Grid>
```


src/lib/components/Motif/Pickers/components/PickerActions.tsx
```diff
@@ -2,23 +2,26 @@ import styles from "../Picker.module.scss";
...
 import { Size4SM } from "../../../../types";
 import { sanitizeModuleClasses } from "../../../../../utils/cssUtils";
+import { useMotifContext } from "src/lib/motif/context/MotifProvider.tsx";
 
 type Props = {
   size: Size4SM;
   onClearClick?: () => void;
   onOkClick?: () => void;
   spread?: boolean;
+  locale?: unknown;
```
ZehranurC: It looks like the locale prop isn't being used anywhere. I'd suggest removing it to keep the code clean.

src/i18n/translate.ts
```diff
@@ -2,28 +2,37 @@ import { Locale, locales } from "./locales/index.ts";
...
   (locale: Locale, localeTexts?: DeepPartial<LocaleShape>): LibraryTranslateFn =>
   (key: LocaleKey, params?: Record<string, unknown>): string => {
     const template =
-      (localeTexts && getNestedValue(localeTexts, key)) ?? getNestedValue(locales[locale], key) ?? getNestedValue(locales.en, key) ?? key;
-    return interpolate(template, params);
+      (localeTexts && getNestedValue(localeTexts, key as string)) ??
+      getNestedValue(locales[locale], key as string) ??
+      getNestedValue(locales.en, key as string) ??
+      key;
+    return <string>interpolate(template as string, params);
```
ZehranurC: `LibraryTranslateFn` type declares return type as string but `interpolate()` can return string | string[]. Then string cast is applied.

Are you sure this is correct?

atakanbayrak: Very nice catch, I built this code so idiotically. Next commit will introduce new createTranslator method for solution to support array and single string.

src/i18n/locales/tr.json
```diff
@@ -62,5 +62,20 @@
   },
   "misc": {
     "playgroundDescription": "This is the playground section to try your components..."
+  },
+  "datepicker": {
```
aktasmehmet: ``date``would be a better naming instead of ``datepicker``. This is not only related with datepicker. This may lead to confusion

src/i18n/locales/tr.json
```diff
@@ -62,5 +62,20 @@
...
+  },
+  "datepicker": {
+    "months": ["Ocak", "Şubat", "Mart", "Nisan", "Mayıs", "Haziran", "Temmuz", "Ağustos", "Eylül", "Ekim", "Kasım", "Aralık"],
+    "monthsShort": ["Oca", "Şub", "Mar", "Nis", "May", "Haz", "Tem", "Ağu", "Eyl", "Eki", "Kas", "Ara"],
+    "weekDays": ["Pz", "Pt", "Sa", "Çr", "Pe", "Cu", "Ct"],
+    "firstDayOfWeek": 1,
+    "last": "Son",
+    "days": "gün",
+    "today": "Bugün",
+    "choose": "Lütfen seçiniz",
```
aktasmehmet: ``choose`` is very generic. We should move this into ``g``

src/lib/components/DatePicker/components/DateSelector/MonthYear/MonthSelector.tsx
```diff
@@ -1,13 +1,18 @@
...
 import MonthYearButton from "./MonthYearButton";
+import { useMotifContext } from "src/lib/motif/context/MotifProvider.tsx";
+import { getDateLocale } from "src/i18n/locales/dateLocals.ts";
 
 const MonthSelector = () => {
-  const { pickerDate, setPickerDate, setPicker, locale, onPickerChange, size } = useContext(DatePickerContext);
+  const { pickerDate, setPickerDate, setPicker, locale: contextLocale, onPickerChange, size } = useContext(DatePickerContext);
+
+  const { t } = useMotifContext();
+  const locale = useMemo(() => contextLocale ?? getDateLocale(t), [contextLocale, t]);
```
aktasmehmet: You've used the same pattern in a lot of places. I am thinknig that maybe creating another hook called ``useDateLocale`` with a generic type and get the locale from that?

src/lib/components/DatePicker/types.ts
```diff
@@ -46,7 +45,7 @@ export type DatePickerPickerType = "day" | "month" | "year";
 
 export type DatePickerContextProps = {
   size: Size4SM;
-  locale: DatePickerLocale;
+  locale?: DatePickerLocale;
```
aktasmehmet: according to your code in DatePicker.tsx, this should not be optional. I know you did this because of the default value thing below but you have to find another solution there. Maybe in the context, the type should be ``null | DatePickerContextProps`` instead of ``DatePickerContextProps`` and you may give null as default value?

src/lib/components/DateRangePicker/types.ts
```diff
@@ -52,7 +51,7 @@ export type DateRangePickerProviderProps = {
 };
 
 export type DateRangePickerContextProps = {
-  locale: DateRangePickerLocale;
+  locale?: DateRangePickerLocale;
```
aktasmehmet: I've written a comment related with this. This is applicable for all of the same stuff in the code. This is just a reminder, you may resolve this.

src/lib/components/TimePicker/components/TimeSelector/TimeHeader.tsx
```diff
@@ -1,9 +1,14 @@
...
 import { TimePickerContext } from "@/components/TimePicker/context/TimePickerProvider";
+import { useMotifContext } from "src/lib/motif/context/MotifProvider.tsx";
+import { getDateLocale } from "src/i18n/locales/dateLocals.ts";
 
 const TimeHeader = () => {
-  const { locale, secondsEnabled } = useContext(TimePickerContext);
+  const { locale: contextLocale, secondsEnabled } = useContext(TimePickerContext);
+
+  const { t } = useMotifContext();
+  const locale = useMemo(() => contextLocale ?? getDateLocale(t), [contextLocale, t]);
```
aktasmehmet: Why dont you use this line to set contextLocale at the root of the component instead of doing the same thing in sub components? You dont have to change anything here at all.

This i also applicable for other components as well.

src/lib/components/DatePicker/components/DateSelector/Day/DayHeader.tsx
```diff
@@ -15,7 +15,7 @@ const DayHeader = () => {
 
   return (
     <Header
-      month={locale.months[pickerDate.getMonth()]}
+      month={locale!.months[pickerDate.getMonth()]}
```
aktasmehmet: I believe you can overcome this kind of situations

src/i18n/locales/dateLocals.ts
```diff
@@ -0,0 +1,20 @@
+import { DaysOfWeek } from "@/components/Motif/Pickers/types";
+import { LibraryTranslateFn } from "src/i18n/types.ts";
+
+export const getDateLocale = (t: LibraryTranslateFn) => ({
```
aktasmehmet: this is a helper function annd ``/locales`` folder is there to only contain the locale files. So:

- You should move this file one level up
- I would rename that as helper.ts

src/lib/components/DatePicker/DatePicker.test.tsx
```diff
@@ -294,18 +295,18 @@ describe("DatePicker", () => {
   it("should not render clear and ok buttons when removeActionButtons is set to true", () => {
     const { rerender, queryByText } = render(<DatePicker />);
     expect(queryByText("Clear")).toBeInTheDocument();
-    expect(queryByText("OK")).toBeInTheDocument();
+    expect(queryByText("Submit")).toBeInTheDocument();
```
aktasmehmet: "Submit" is the wrong value here. It is not submitting anything since it is not a form. "OK" should stay and this should be same for all other files.

src/lib/components/Motif/Pickers/components/PickerActions.tsx
aktasmehmet: 💯

src/i18n/locales/tr.json
```diff
@@ -62,5 +62,20 @@
   },
   "misc": {
     "playgroundDescription": "This is the playground section to try your components..."
+  },
+  "datepicker": {
+    "months": ["Ocak", "Şubat", "Mart", "Nisan", "Mayıs", "Haziran", "Temmuz", "Ağustos", "Eylül", "Ekim", "Kasım", "Aralık"],
+    "monthsShort": ["Oca", "Şub", "Mar", "Nis", "May", "Haz", "Tem", "Ağu", "Eyl", "Eki", "Kas", "Ara"],
+    "weekDays": ["Pz", "Pt", "Sa", "Çr", "Pe", "Cu", "Ct"],
+    "firstDayOfWeek": 1,
```
aktasmehmet: I dont think this is good. We should keep the balance between easy to integrate vs consistency. There should not be a number in a tranlation file.

There is nothing comes to my mind right now but you have to figure this out.

src/i18n/types.ts
```diff
@@ -32,4 +36,7 @@ export type LocaleKey = DeepKeys<typeof locales.en>;
  * The signature is intentionally compatible with i18next's `t` so consumers
  * can pass their own t to <MotifProvider t={t}> without type errors.
  */
-export type LibraryTranslateFn = (key: LocaleKey, params?: Record<string, unknown>) => string;
+export type LibraryTranslateFn = {
+  (key: LocaleKey, params?: Record<string, unknown>): string;
+  array(key: LocaleKey, params?: Record<string, unknown>): string[];
+};
```
aktasmehmet: Can you double chech this change? It looks like it is redundant.

src/i18n/types.ts
```diff
@@ -17,10 +17,14 @@ export type LocaleConfig = Locale | { locale?: Locale; texts?: DeepPartial<Local
 type DeepKeys<T, Prefix extends string = ""> = {
   [K in keyof T & string]: T[K] extends string
     ? `${Prefix}${K}`
-    : T[K] extends Record<string, unknown>
-      ? `${Prefix}${K}` | DeepKeys<T[K], `${Prefix}${K}.`>
-      : never;
-}[keyof T & string];
+    : T[K] extends number
```
aktasmehmet: I've mentioned in some other place that number should not be in the translation texts, so this also should not be here. Can you please also double check the situation here for the arrays 🙏

src/i18n/translate.ts
```diff
@@ -2,28 +2,41 @@ import { Locale, locales } from "./locales/index.ts";
...
+const getNestedValue = (obj: Record<string, unknown>, key: string): string | string[] | undefined => {
   const parts = key.split(".");
   let current: unknown = obj;
   for (const part of parts) {
     if (typeof current !== "object" || current === null) return undefined;
     current = (current as Record<string, unknown>)[part];
   }
-  return typeof current === "string" ? current : undefined;
+  if (typeof current === "string") return current;
+  if (typeof current === "number") return String(current);
```
aktasmehmet: number should not be a result for a translation

src/i18n/types.ts
aktasmehmet: Besides, this should be same as the i18next type to support other i18n libraries. Please check the text below in docs:

``The expected signature is intentionally compatible with i18next:``

src/i18n/translate.ts
```diff
@@ -2,28 +2,41 @@ import { Locale, locales } from "./locales/index.ts";
...
+export const createTranslator = (locale: Locale, localeTexts?: DeepPartial<LocaleShape>): LibraryTranslateFn => {
+  const resolve = (key: LocaleKey, params?: Record<string, unknown>) => {
     const template =
-      (localeTexts && getNestedValue(localeTexts, key)) ?? getNestedValue(locales[locale], key) ?? getNestedValue(locales.en, key) ?? key;
-    return interpolate(template, params);
+      (localeTexts && getNestedValue(localeTexts, key as string)) ??
+      getNestedValue(locales[locale], key as string) ??
+      getNestedValue(locales.en, key as string) ??
+      key;
+    return interpolate(template as string, params);
```
aktasmehmet: this is a blind cast, which may result some issues.

since, ``getNestedValues`` can now return an array instead of string, template might be an array of strings?

src/i18n/translate.ts
```diff
@@ -2,28 +2,41 @@ import { Locale, locales } from "./locales/index.ts";
...
-  !params
-    ? template
-    : template.replace(/{{(\w+)}}/g, (_, key: string) => {
-        const val = params[key];
-        return typeof val === "string" || typeof val === "number" ? String(val) : `{{${key}}}`;
-      });
+const interpolate = (template: string | string[], params?: Record<string, unknown>): string | string[] => {
+  if (Array.isArray(template) || !params) {
+    return template;
+  }
```
aktasmehmet, line 20: I dont think it is necessary right now but with this short circuit, you basically prevent translations like:

```
["Welcome {{name}}", "Enjoy your stay"], {{name}}
```

src/i18n/translate.ts
```diff
@@ -2,28 +2,41 @@ import { Locale, locales } from "./locales/index.ts";
...
-    return interpolate(template, params);
+      (localeTexts && getNestedValue(localeTexts, key as string)) ??
+      getNestedValue(locales[locale], key as string) ??
+      getNestedValue(locales.en, key as string) ??
+      key;
+    return interpolate(template as string, params);
   };
+
+  const t = (key: LocaleKey, params?: Record<string, unknown>): string => resolve(key, params) as string;
+  t.array = (key: LocaleKey, params?: Record<string, unknown>): string[] => resolve(key, params) as string[];
```
aktasmehmet: I dont actually understand here, the array prop override stuff. I cant find any usage of ``.array`` in any ``t`` usage at all.

Besides, I think this lines may cause issues like:

- If user calls like t("some.array.key"), this will return an array but typescript will behave like it is a string. For example if developer will call toLowercase or something, app will crush in the run time. Exact opposite scenario may also happen with array<->string

aktasmehmet (commented):
Thx, good work. I think that especially ``translate.ts`` should be re-checked 😉

src/i18n/translate.ts
atakanbayrak: This only remains, I will look into it deeply when my mind is clear.

src/i18n/locales/en.json
```diff
@@ -14,7 +14,9 @@
     "previous": "Previous",
     "next": "Next",
     "finish": "Finish",
-    "x_message": "{{count}} message"
+    "x_message": "{{count}} message",
+    "ok": "OK",
+    "choose": "Please choose"
```
aktasmehmet: I would rename this as "choosePlease" to prevent a possible key chain work in the future when ``choose`` word might be used standalone.

src/lib/hooks/useDateLocale.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * @param contextLocale - An optional override locale string or object
+ * @returns The resolved date locale
+ */
+export function useDateLocale<T = DateLocale>(contextLocale?: T): T | DateLocale {
+  const { t } = useMotifContext();
+
+  return useMemo(() => {
+    return contextLocale ?? getDateLocale(t);
+  }, [contextLocale, t]);
```
aktasmehmet:
```suggestion
  return useMemo(() => contextLocale ?? getDateLocale(t), [contextLocale, t]);
```

src/i18n/useDateLocale.ts
```diff
@@ -0,0 +1,18 @@
+import { useMemo } from "react";
```
aktasmehmet, line 1: We put commonly usable hooks under this folder. This hook has a more specific context. So, you may consider moving this somewhere more date ot locale related.

src/lib/hooks/useDateLocale.ts
```diff
@@ -0,0 +1,18 @@
...
+import { DateLocale, getDateLocale } from "src/i18n/helper.ts";
+
+/**
+ * A custom hook to fetch the current date locale.
+ * It prioritizes an explicitly passed locale, falling back to the motif context.
+ *
+ * @param contextLocale - An optional override locale string or object
+ * @returns The resolved date locale
+ */
+export function useDateLocale<T = DateLocale>(contextLocale?: T): T | DateLocale {
```
aktasmehmet:
```suggestion
export function useDateLocale<T = DateLocale>(contextLocale?: T) {
```

You dont need to explicitly set the return type. It will be calculated automatically by typescript.

src/lib/components/TimePicker/TimePicker.tsx
```diff
@@ -15,7 +15,7 @@ const TimePicker = (props: PropsWithRef<TimePickerProps, HTMLDivElement>) => {
     fluid,
     onOkClick,
     onPeriodChange,
-    locale = LOCALE_TIME_PICKER_TR_TR,
+    locale,
```
aktasmehmet: in InputTime and InputDate, you've renamed this as ``propsLocale``and used ``locale`` instead of ``resolvedLocale`` in the render part.

This code is not wrong but you may consider doing the same here to keep it more consistent?

src/lib/components/DatePicker/DatePicker.tsx
```diff
@@ -16,22 +16,24 @@ export const DatePicker = (props: PropsWithRef<DatePickerProps, HTMLDivElement>)
     fluid,
     onOkClick,
     value,
-    locale = LOCALE_DATE_TR_TR,
+    locale,
     onDateChange,
     onPickerChange,
     onClearClick,
     removeActionButtons,
+    firstDayOfWeek = 1,
```
aktasmehmet, line 24: This default value should also be added to the storybook. Please apply the same for other components as well

src/lib/components/DatePicker/DatePicker.test.tsx
aktasmehmet: Unfortunately, we should split this from locale test and write its own test for ``firstDayOfWeek`` props. You should do the same for all other components as well

src/lib/components/DateRangePicker/types.ts
```diff
@@ -86,4 +84,13 @@ export const datePickerContextDefaultValues: DateRangePickerContextProps = {
...
   getDaysOfMonth: () => [],
   initialMonths: [],
+  locale: {
+    months: [],
+    monthsShort: [],
+    weekDays: [],
+    today: "",
+    last: "",
+    days: "",
+    choose: "",
```
aktasmehmet, line 92: **choose** should not be here I guess

src/i18n/helper.ts
```diff
@@ -0,0 +1,33 @@
+import { LibraryTranslateFn } from "src/i18n/types.ts";
+
+export type DateLocale = {
```
aktasmehmet: I would move this type to types.ts

src/lib/components/DateRangePicker/types.ts
aktasmehmet: firstDayOfWeek should be removed here

src/i18n/helper.ts
```diff
@@ -0,0 +1,33 @@
...
+  months: string[];
+  monthsShort: string[];
+  weekDays: string[];
+  today: string;
+  last: string;
+  days: string;
+  choose: string;
```
aktasmehmet: ``choose``s in this file should be removed I guess?

src/i18n/helper.ts
```diff
@@ -0,0 +1,33 @@
...
+  hoursAbbr: string;
+  minutesAbbr: string;
+  secondsAbbr: string;
+  am: string;
+  pm: string;
+};
+
+export const getDateLocale = (t: LibraryTranslateFn): DateLocale => ({
+  months: t("date.months") as unknown as string[],
```
aktasmehmet, line 4: There is an issue here we need to discuss. You are casting this to string[] because it has to be like this, I know.

But, ``t()``function is used here and it should be i18next compatible to use custom translators, right? So, i18next compatible t functions always return string. So, if a user would use a custom translator, these lines will probably throw errors.

If this is too much to handle here, you may create a custom task for this but we should definitely think over and solve this problem.

aktasmehmet (changes requested):

src/lib/components/DateRangePicker/types.ts
atakanbayrak: DateRangePickerLocale value has it so I add it on.

src/i18n/helper.ts
atakanbayrak: Nope, they are necessary to understand if locale value comes from prop or library locale. If values will be selected from library locale we have to have this value to set it on inside the component.

src/i18n/helper.ts
atakanbayrak: I'll open a task for it and also same one with that you already commented out https://github.com/motif-ui/motifreact/pull/81#discussion_r3450710097.

They are both work together so any changes on the other comment will be outdated when we discuss this problem.

src/i18n/translate.ts
atakanbayrak: https://github.com/motif-ui/motifreact/pull/81#discussion_r3457673443

This problem should be solved with the issue commented out in here. I'll open task for it, because I am kind a blind for it right now.

src/lib/components/DateTimePicker/DateTimePicker.test.tsx
```diff
@@ -189,15 +189,14 @@ describe("DateTimePicker", () => {
...
-      firstDayOfWeek: 3 /* 1D */,
       hoursAbbr: "H1",
       minutesAbbr: "M2",
       secondsAbbr: "S3",
       am: "AM",
       pm: "PM",
     };
     const { container, getByText, getAllByText } = render(
-      <DateTimePicker locale={LOCALE_DATETIME_CUSTOM} secondsEnabled timeFormat="12h" />,
+      <DateTimePicker locale={LOCALE_DATETIME_CUSTOM} firstDayOfWeek={3} secondsEnabled timeFormat="12h" />,
```
aktasmehmet: I guess ``firstDayOfWeek`` should be removed here.

src/lib/components/TimePicker/TimePicker.tsx
aktasmehmet: DateTimePicker, DatePicker, DateRangePicker still has the same usage

ZehranurC (dismissed):

aktasmehmet (dismissed):

aktasmehmet (approved):

PR #82, opened 2026-06-08
fix: clearable button feature is added for InputText component

atakanbayrak:
## Description

Clearable button is added for InputText.

## Type of Change

- [ ] 🆕 New Component
- [x] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Screenshots / Preview

Screenshot 2026-06-08 at 16 41 13

## Checklist

## How to Test

npm run test:ci & npm run dev

### DIFF
--- a/src/lib/components/InputText/InputText.test.tsx
+++ b/src/lib/components/InputText/InputText.test.tsx
@@ -15,6 +15,11 @@ describe("InputText", () => {
     expect(screen.getByDisplayValue(value));
   });
 
+  it("should display clearable icon when clearable prop is given", () => {
+    render(<InputText clearable />);
+    expect(screen.getByTestId("iconButtonTestId")).toBeInTheDocument();
+  });
+
   it("should be rendered with the size given in size prop", () => {
     const sizes: InputSize[] = ["xs", "sm", "md", "lg"];
     for (const size of sizes) {
--- a/src/lib/components/InputText/InputText.test.tsx
+++ b/src/lib/components/InputText/InputText.test.tsx
@@ -15,9 +15,15 @@ describe("InputText", () => {
     expect(screen.getByDisplayValue(value));
   });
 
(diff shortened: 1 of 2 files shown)


src/lib/components/InputText/InputText.test.tsx
```diff
@@ -15,6 +15,11 @@ describe("InputText", () => {
     expect(screen.getByDisplayValue(value));
   });
 
+  it("should display clearable icon when clearable prop is given", () => {
```
aktasmehmet:
```suggestion
  it("should display clearable icon when clearable prop is true", () => {
```

src/lib/components/InputText/InputText.test.tsx
```diff
@@ -15,6 +15,11 @@ describe("InputText", () => {
...
+    render(<InputText clearable />);
+    expect(screen.getByTestId("iconButtonTestId")).toBeInTheDocument();
+  });
+
```
aktasmehmet, line 22: I think you should are more tests to cover like:

- clears when clicked
- only appears when there is a text,
- etc.

ZehranurC: +1

atakanbayrak: Clear icon appearing is not related to existence of text. When prop is given, clearable icon will be there either. So I added case for clearing text.

ZehranurC (approved):

ZehranurC:
🚀

src/lib/components/InputText/InputText.test.tsx
aktasmehmet: This is actually a bug and an issue is there for that, I've checked. Thx

aktasmehmet (approved):

PR #84, opened 2026-06-09
fix button

akhfaern:
## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Checklist

### DIFF
--- a/src/lib/components/Button/Button.module.scss
+++ b/src/lib/components/Button/Button.module.scss
@@ -59,11 +59,8 @@ $variants: (
 .fluid {
   width: 100%;
 }
-.icon-left {
-  flex-direction: row;
-}
-.icon-right {
-  flex-direction: row-reverse;
+.icon-only {
+  gap: 0;
 }
 
 @each $size, $ph, $pv, $fontSize, $iconSize in $sizes {
@@ -71,8 +68,13 @@ $variants: (
     padding: var(--base-sizing-#{$pv}) var(--base-sizing-#{$ph});
     gap: var(--base-sizing-#{$pv});
     @include font("base", $fontSize, "600");
+
     .icon {
       --mtf-icon-size: var(--base-sizing-#{$iconSize});
+    }
+
+    &.icon-only {
+      padding: var(--base-sizing-#{$pv});
     }
   }
 }
--- a/src/lib/components/Button/Button.tsx
+++ b/src/lib/components/Button/Button.tsx
@@ -1,15 +1,19 @@
 "use client";
 
+import { useEffect, type ReactNode } from "react";
 import styles from "./Button.module.scss";
 import GlobalIconWrapper from "../Motif/GlobalIconWrapper/GlobalIconWrapper";
-import { PropsWithRef } from "../../types";
 import { sanitizeModuleRootClasses } from "../../../utils/cssUtils";
 import usePropsWithThemeDefaults from "../../motif/hooks/usePropsWithThemeDefaults";
 import type { ButtonProps } from "./types";
 
-const Button = (props: PropsWithRef<ButtonProps, HTMLButtonElement>) => {
+const hasRenderableContent = (value: ReactNode) =>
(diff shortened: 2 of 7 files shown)


src/lib/components/Button/Button.tsx
```diff
@@ -18,31 +22,58 @@ const Button = (props: PropsWithRef<ButtonProps, HTMLButtonElement>) => {
...
 
+  const content = children ?? label;
+  const hasContent = hasRenderableContent(content);
+  const isIconOnly = Boolean(icon && !hasContent);
+
+  useEffect(() => {
+    if (process.env.NODE_ENV !== "production" && isIconOnly && !ariaLabel) {
+      console.warn("Button: Icon-only buttons must provide an accessible name with `aria-label`.");
+    }
+  }, [isIconOnly, ariaLabel]);
```
atakanbayrak, line 41: Checking if this process works under production or development environment inside component is not best case, because starting to create these useEffects inside components will cause spagetti code in the future to make it harder to understand components logic.

src/lib/components/Button/Button.module.scss
```diff
@@ -59,21 +59,23 @@ $variants: (
 .fluid {
   width: 100%;
 }
-.icon-left {
-  flex-direction: row;
-}
-.icon-right {
-  flex-direction: row-reverse;
```
atakanbayrak, line 66: With the deletion of iconLeft and iconRight, we lost the feature of handling icon positioning inside component.

src/lib/components/Button/types.ts
```diff
@@ -1,14 +1,15 @@
...
 export type ButtonProps = {
-  label?: string;
+  ref?: Ref<HTMLButtonElement>;
+  label?: ReactNode;
+  children?: ReactNode;
   icon?: IconGlobalType;
   iconPosition?: "left" | "right";
-  disabled?: boolean;
-  onClick?: (event: MouseEvent<HTMLButtonElement>) => void;
-  htmlType?: "submit" | "button";
```
atakanbayrak, line 10: I understand that, icon only mode could not require disabled or onClick or htmlType event, but we will need these props if we use it without iconOnly mode and iconOnly mode either.

src/lib/components/Button/types.ts
```diff
@@ -1,14 +1,15 @@
 import type { IconGlobalType } from "../../types";
-import type { MouseEvent } from "react";
+import type { ButtonHTMLAttributes, ReactNode, Ref } from "react";
 
 export type ButtonProps = {
-  label?: string;
```
atakanbayrak, line 5: Deletion of label is not possible for Button component. It will serve us to use it only icon-mode in advance.

atakanbayrak (changes requested):
First of all, thanks for the contribution and spending your time to maintain our button component. We welcome warmly your button component fix and happy to review it. There are several issues I would like to talk in advance. Tell me if I am wrong, but you add icon-only mode feature to component while reverting previous features, with using left or right icon mode and disabled etc. Please consider find a solution with adding icon-only mode to component without affecting previous features.

akhfaern:
:D not worth of my time

src/lib/components/Button/Button.tsx
akhfaern: please learn what is best practice what is not

src/lib/components/Button/Button.module.scss
akhfaern: no you dont

src/lib/components/Button/types.ts
akhfaern: these come from the basic types you don't need to re define them. learn basic types

src/lib/components/Button/types.ts
akhfaern: i did not delete it i add line before label please learn to read code

PR #85, opened 2026-06-10
fix: logo prop refactor to component based

ZehranurC:
## Description

NavBar logo prop refactored to accept React components instead of image paths.

**Changes:**
- Removed `imgPath: string` prop
- Added `image: ReactElement` for flexible component rendering
- Supports ``, `<Image>`, `<NextImage>`, and other custom components

## Type of Change

- [ ] 🆕 New Component
- [x] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [x] 💥 Breaking Change
- [x] 🧹 Code Refactoring

## Checklist

## How to Test

```jsx
<NavBar logo={{ image: <img src="logo.png" alt="Logo" />, href: "/" }} />
```


src/lib/components/NavBar/NavBar.test.tsx
```diff
@@ -181,6 +181,17 @@ describe("NavBar", () => {
...
+    const CustomLogo = ({ title }: { title: string }) => (
+      <svg role="img" aria-label={title}>
+        <title>{title}</title>
+      </svg>
+    );
+    const { getByRole } = render(<NavBar logo={{ image: <CustomLogo title="Custom logo" /> }} />);
+
+    expect(getByRole("img", { name: "Custom logo" })).toBeInTheDocument();
+  });
+
```
atakanbayrak: What I understand that there is no alt message support if user provided none. At least we could define default value for alt prop to show some value if there is a problem while rendering custom image or logo.

atakanbayrak (changes requested):

src/lib/components/NavBar/components/NavBarLogo.tsx
```diff
@@ -1,19 +1,19 @@
+import type { ReactElement } from "react";
 import styles from "../NavBar.module.scss";
 
 export type NavBarLogoProps = {
-  imgPath: string;
-  alt?: string;
+  image: ReactElement;
   href?: string;
```
aktasmehmet, line 6: Actually this change got me stop reviewing any further. This is an additional feature and we dont need to drop supporting previous props. If you do this way, the apps using this props will collapse.

So, I would keep the current props as it is and still render an , put an additional ``logoSlot: ReactElement`` prop to the Navbar itself. if logo is given, it should render that. If only logoSlot is given, then render that.

aktasmehmet (changes requested):

src/lib/components/NavBar/NavBar.test.tsx
ZehranurC: With the new commit, we've moved to a flexible approach: `logoSlot: ReactElement`
This allows users to pass any component (img, Image, SVG, etc.) 
with full control over attributes like alt etc.

This resolves the concern about alt text support.

src/lib/components/NavBar/components/NavBarLogo.tsx
ZehranurC: Fixed in the latest commit. You can verify it.

src/lib/components/NavBar/NavBar.tsx
```diff
@@ -35,7 +36,7 @@ const NavBar = (props: PropsWithRef<NavBarProps, HTMLElement>) => {
         {mainMenu && (
           <MotifIconButton name="density_medium" size="xl" onClick={() => setMenuCollapsed(p => !p)} className={styles.hamburger} />
         )}
-        {logo && <NavBarLogo {...logo} />}
+        {logoSlot ? <NavBarLogo image={logoSlot} /> : logo ? <NavBarLogo {...logo} /> : null}
```
aktasmehmet:
```suggestion
        {(logo || logoSlot) && <NavBarLogo image={logoSlot} {...logo} />}
```

src/lib/components/NavBar/components/NavBarLogo.tsx
```diff
@@ -1,19 +1,23 @@
...
+  imgPath?: string;
   alt?: string;
   href?: string;
+  image?: ReactElement;
 };
 
 const NavBarLogo = (props: NavBarLogoProps) => {
-  const { href, imgPath, alt } = props;
+  const { href, imgPath, alt, image } = props;
+  const logoContent = image ? image : <img src={imgPath} alt={alt} />;
```
aktasmehmet:
```suggestion
  const logoContent = image ?? <img src={imgPath} alt={alt} />;
```

src/lib/components/NavBar/components/NavBarLogo.tsx
```diff
@@ -1,19 +1,23 @@
...
+
   return href ? (
     <a href={href} className={styles.logo}>
-      <img src={imgPath} alt={alt} />
+      {logoContent}
     </a>
   ) : (
-    <img src={imgPath} alt={alt} className={styles.logo} />
+    <div className={styles.logo}>{logoContent}</div>
```
aktasmehmet: 1 - In the documentation of types.ts, you've mentioned the **precedence** of logoSlot prop. So, you should also update here to first check ``image``, then ``href``
2 - More importantly, now, logo class becomes a container. It is not feasible to use the same class with an ``<a>`` and some unknown component or div. So I would use a div to wrap everything, and conditionally render anchor or the element givn by the user here

src/lib/components/NavBar/NavBar.module.scss
```diff
@@ -24,8 +24,12 @@ $variants: ("neutral", "primary", "secondary", "info", "danger", "warning", "suc
 }
 
 .logo {
+  display: flex;
+  align-items: center;
   height: var(--base-sizing-40x);
-  img {
+
+  > *,
+  > * > * {
```
aktasmehmet: not a fan of this usage :) I'll check again after the changes on the wrapper side. Probably there will be no need here.

src/lib/components/NavBar/types.ts
```diff
@@ -79,6 +79,12 @@ export type NavBarDefaultableProps = {
    * ```
    */
   logo?: NavBarLogoProps;
+  /**
+   * Custom logo component. Takes precedence over logo prop.
+   * Use for flexible logo rendering with <img>, <Image>, <NextImage>, etc.
+   * Wrap with <a> tag if you need a link: <a href="/home"><img .../></a>
```
aktasmehmet: First line is enough I guess. Remaining part is redundant.


```suggestion
```

src/lib/components/NavBar/NavBar.test.tsx
```diff
@@ -181,6 +181,42 @@ describe("NavBar", () => {
...
+      <svg role="img" aria-label={title}>
+        <title>{title}</title>
+      </svg>
+    );
+    const { getByRole } = render(<NavBar logoSlot={<CustomLogo title="Custom SVG Logo" />} />);
+
+    expect(getByRole("img", { name: "Custom SVG Logo" })).toBeInTheDocument();
+  });
+
+  it("should render logoSlot with href", () => {
```
aktasmehmet: This test is not necessary. User can use any element, now just anchor. So, no need here.

Having said that, you may consider moving the content of this test to ``should render custom component using logoSlot prop`` test, instead of svg there.

src/lib/components/NavBar/NavBar.test.tsx
aktasmehmet: I think this test is redundant right now?

src/lib/components/NavBar/types.ts
```diff
@@ -79,6 +79,12 @@ export type NavBarDefaultableProps = {
    * ```
    */
   logo?: NavBarLogoProps;
+  /**
+   * Custom logo component. Takes precedence over logo prop.
```
aktasmehmet, line 84: You should add a test for this.

aktasmehmet (changes requested):

src/lib/components/NavBar/types.ts
aktasmehmet: just a single line of explanation is enoguh I guess. "Use for with..." is not necessary because the type if ReactElement, so anything can be used here. We just need ``Custom logo component. Takes precedence over logo prop.``, remaining is redundant.

src/lib/components/NavBar/NavBar.test.tsx
```diff
@@ -213,10 +191,19 @@ describe("NavBar", () => {
...
     expect(getByRole("link")).toHaveAttribute("href", "https://example.com/");
     expect(getByRole("img")).toHaveAttribute("src", "logo.png");
   });
 
+  it("should prioritize logoSlot over logo prop", () => {
+    const { getByRole } = render(
+      <NavBar logo={{ image: <img src="logo-prop.png" alt="Logo prop" /> }} logoSlot={<img src="logo-slot.png" alt="Logo slot" />} />,
+    );
+
+    expect(getByRole("img")).toHaveAttribute("src", "logo-slot.png");
```
aktasmehmet, line 212: can you please add an "expect not" for the logo prop here?

src/lib/components/NavBar/NavBar.test.tsx
aktasmehmet: @atakanbayrak ?

src/lib/components/NavBar/components/NavBarLogo.tsx
```diff
@@ -1,19 +1,17 @@
+import type { ReactElement } from "react";
 import styles from "../NavBar.module.scss";
 
 export type NavBarLogoProps = {
-  imgPath: string;
+  imgPath?: string;
   alt?: string;
   href?: string;
+  image?: ReactElement;
```
aktasmehmet: Why did you add this here? With this, you've just reverted back to the start where we said that this prop should not change and an additional logoSlot prop should handle the rest.

There is logoSlot now. If the user want to put an image, that prop can be used. So, this would be a "make-it-easy" prop and generate an  element, not get one.

So, ``image`` prop should be removed here

src/lib/components/NavBar/NavBar.test.tsx
```diff
@@ -172,7 +172,7 @@ describe("NavBar", () => {
     const { getByRole } = render(
       <NavBar
         logo={{
-          imgPath: "some_image",
+          image: <img src="some_image" alt="logo" />,
```
aktasmehmet: logo should not have an image prop

src/lib/components/NavBar/types.ts
```diff
@@ -79,6 +79,11 @@ export type NavBarDefaultableProps = {
...
+   * Use for flexible logo rendering with <img>, <Image>, <NextImage>, etc.
+   */
+  logoSlot?: React.ReactElement;
```
aktasmehmet: "React." part should be moved to the "import" section.

src/lib/components/NavBar/NavBar.tsx
```diff
@@ -35,7 +36,7 @@ const NavBar = (props: PropsWithRef<NavBarProps, HTMLElement>) => {
         {mainMenu && (
           <MotifIconButton name="density_medium" size="xl" onClick={() => setMenuCollapsed(p => !p)} className={styles.hamburger} />
         )}
-        {logo && <NavBarLogo {...logo} />}
+        {(logo || logoSlot) && <NavBarLogo {...logo} image={logoSlot || logo?.image} />}
```
aktasmehmet: I might suggest an easier solution here:

You may consider getting ``logoSlot`` as children for ``<NavBarLogo>``and make this like:


```suggestion
        {(logo || logoSlot) && <NavBarLogo {...logo}>{logoSlot}</NavBarLogo>}
```

Then in NavBarLogo, check children... If there is children, then render that. If not, then generate **** with props coming from ``logo``

src/lib/components/NavBar/components/NavBarLogo.tsx
```diff
@@ -1,19 +1,17 @@
...
-  const { href, imgPath, alt } = props;
-  return href ? (
-    <a href={href} className={styles.logo}>
-      <img src={imgPath} alt={alt} />
-    </a>
-  ) : (
-    <img src={imgPath} alt={alt} className={styles.logo} />
-  );
+  const { href, imgPath, alt, image } = props;
+  const logoContent = image ? href ? <a href={href}>{image}</a> : image : <img src={imgPath} alt={alt} />;
```
aktasmehmet: I've put a comment include as change here, in ``NavBar.tsx``

aktasmehmet (changes requested):

src/lib/components/NavBar/components/NavBarLogo.tsx
ZehranurC: Fixed it with new commit

src/lib/components/NavBar/NavBar.tsx
ZehranurC: Fixed it with new commit

src/lib/components/NavBar/components/NavBarLogo.tsx
```diff
@@ -1,19 +1,25 @@
+import { PropsWithChildren } from "react";
 import styles from "../NavBar.module.scss";
 
 export type NavBarLogoProps = {
-  imgPath: string;
+  imgPath?: string;
```
aktasmehmet: If logo prop is given, imgPath should be there. So, this should not be optional

src/lib/components/NavBar/components/NavBarLogo.tsx
```diff
@@ -1,19 +1,25 @@
...
+    (imgPath &&
+      (href ? (
+        <a href={href}>
+          <img src={imgPath} alt={alt} />
+        </a>
+      ) : (
+        <img src={imgPath} alt={alt} />
+      )));
+
+  return logoContent ? <div className={styles.logo}>{logoContent}</div> : null;
```
aktasmehmet: You already check ``logo || logoSlot``before calling this component. So, no need for a second check here:


```suggestion
const NavBarLogo = (props: PropsWithChildren<NavBarLogoProps>) => {
  const { children, imgPath, alt, href } = props;
  return (
    <div className={styles.logo}>
      {children ??
        (href ? (
          <a href={href}>
            <img src={imgPath} alt={alt} />
          </a>
        ) : (
          <img src={imgPath} alt={alt} />
        ))}
    </div>
  );
```

src/lib/components/NavBar/NavBar.tsx
```diff
@@ -35,7 +36,7 @@ const NavBar = (props: PropsWithRef<NavBarProps, HTMLElement>) => {
         {mainMenu && (
           <MotifIconButton name="density_medium" size="xl" onClick={() => setMenuCollapsed(p => !p)} className={styles.hamburger} />
         )}
-        {logo && <NavBarLogo {...logo} />}
+        {(logo || logoSlot) && <NavBarLogo {...logo}>{logoSlot}</NavBarLogo>}
```
aktasmehmet: I've directly tested the component in primary story and here is the result:

Image

Logo is not rendered

aktasmehmet (changes requested):

aktasmehmet (approved):

atakanbayrak (approved):

PR #89, opened 2026-06-18
fix(Upload): value prop is added

aktasmehmet:
## Description

This PR basically integrates files already in the server, into the Upload components (UploadDragger, List, Input) with the value prop. In order to achieve this, a new external type is introduced and components are fixed to map that type to the internal file types. Form side registrations are also enhanced.

I tried to gather together the changes I've made and issues I've faced below. They are not all of the changes but at least this gives an insight.  

| Issue | Severity | Detail |
  |---|---|---|
  | `uploaded: true` missing | Critical | Added to initializer; guards in validation, removeFiles, and maxFile checks all key off `f.uploaded` |
  | STATUS.UPLOADED → reuse SUCCESS | Design | Reverted new status; server files use STATUS.SUCCESS with `uploaded: true` |
  | Download button gate | High | FileButton checks `file.download` instead of a dedicated status |
  | Validation guard | Critical | `f.uploaded === true` causes early return, skipping size/type/count checks for server files |
  | `removeFiles` routing | High | Server files (`f.uploaded`) go to `_deleteFilesFromServer` instead of local-only removal |
  | maxFile counting | High | Both DragAreas now use `f.status !== STATUS.CHECK_FAIL` to count occupied slots |
  | `_isSameFile` ignores server files | High | Removed `lastModified` from comparison; now matches on `name + size + type` only |
  | `file` is a plain object cast as `File` | Low | `FileType.file` narrowed to `Pick<File, 'name'\|'size'\|'type'>`; `as File` cast at the 3 call sites that need the real instance. Discriminated union/type guard alternatives considered and rejected. |
  | Wasted re-render on mount | Low | Reference equality check before `setSelectedFiles`; server files return same `f` ref from early-return, so no state update fires |
  | Form registered raw `FileObject[]` for UploadList/UploadDragger | Critical | Introduced `mapExternalValue` helper in `helper.ts`; conversion now happens before `useRegisterFormField` so form always receives `FileType[]` |
  | `UploadItemSyncValidation` missing from UploadList/UploadDragger | High | Added `defaultValidations: [Validations.UploadItemSyncValidation]` to both components' `useRegisterFormField` calls |
  | Internal fields leaking into form submit data | Medium | `toFormValue` helper strips `download`, `addedByValue`, and `request` before `onFormFieldValueUpdate` and external `onChange` |
  | Tests for value prop | N/A | 11 tests for UploadDragger, 11 for UploadList, 6 for UploadInput; Form tests for submit data correctness and RequiredUploadedFile validation across all 3 components |

## Type of Change

- [ ] 🆕 New Component
- [x] ✨ Feature / Enhancement
- [x] 🐛 Bug Fix
- [x] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Screenshots / Preview

Screenshot 2026-06-18 at 11 28 15

## Checklist

## How to Test

```
const files: FileObject[] = [
    {
      id: "1",
      name: "file1.png",
      type: "image/png",
      size: 1241234,
      onDownloadClick: () => window.open("https://picsum.photos/200/300", "_blank"),
    },
    {
      id: "2",
      name: "file2.png",
      type: "image/png",
      size: 545242,
      onDownloadClick: () => window.open("https://picsum.photos/200/300", "_blank"),
    },
    {
      id: "3",
      name: "file3.png",
      type: "image/png",
      size: 44325252,
      onDownloadClick: () => window.open("https://picsum.photos/200/300", "_blank"),
    },
  ];

  return (
    <Form onSubmit={console.log} style={{ width: 400, margin: "20 auto" }}>
      <Form.Field name="uploadDragger" label="Upload Dragger" readOnly>
        <UploadDragger uploadRequest={MOCK.uploadRequest} deleteRequest={MOCK.deleteRequest} value={files} maxFile={2} />
      </Form.Field>
      <Form.Field name="uploadList" label="Upload List">
        <UploadList uploadRequest={MOCK.uploadRequest} deleteRequest={MOCK.deleteRequest} value={[files[0]]} maxFile={2} />
      </Form.Field>
      <Form.Field name="uploadList2" label="Upload List Readonly" readOnly>
        <UploadList uploadRequest={MOCK.uploadRequest} deleteRequest={MOCK.deleteRequest} value={files} maxFile={2} />
      </Form.Field>
      <Form.Field name="uploadInput" label="Upload Input">
        <UploadInput uploadRequest={MOCK.uploadRequest} deleteRequest={MOCK.deleteRequest} value={[files[0]]} />
      </Form.Field>
    </Form>
  );
```


src/lib/components/Upload/helper.ts
```diff
@@ -0,0 +1,22 @@
+import { FileObject, FileType } from "@/components/Upload/types.ts";
```
atakanbayrak, line 1: We could add comment line description to this helper to understand what it does at the future.

src/lib/components/Upload/helper.ts
```diff
@@ -0,0 +1,22 @@
...
+        },
+        status: STATUS.SUCCESS,
+        uploaded: true,
+        download: f.onDownloadClick,
+        addedByValue: true,
+      }))
+    : [];
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const toFormValue = (files: FileType[]) => files.map(({ download: _d, addedByValue: _a, request: _r, ...rest }) => rest);
```
atakanbayrak, line 25: Same comment line also could be added for this helper method.

src/lib/components/Upload/UploadDragger/UploadDragger.test.tsx
```diff
@@ -378,4 +354,74 @@ describe("UploadDragger", () => {
...
+    expect(getFileList()?.childNodes).toHaveLength(2);
+    expect(getFileItemFirst()).toHaveTextContent(serverFile.name);
+    expect(getFileItemLast()).toHaveTextContent(serverFile2.name);
+  });
+
+  it("should send a delete request and remove the file when the delete button is clicked for a value file", async () => {
+    const { getFileList, getDeleteButton } = renderExt(<UploadDragger {...requiredProps} value={[serverFile]} />);
+    await userEvent.click(getDeleteButton());
+    await waitFor(() => expect(getFileList()).not.toBeInTheDocument());
+  });
```
atakanbayrak, line 378: We should add a test like when there is 2 file already uploaded via value prop and deleted one of them, there will be still one file at the end of the process.

src/lib/components/Upload/UploadDragger/UploadDragger.tsx
```diff
@@ -10,6 +10,9 @@ import { InputValue } from "../../Form/types";
 import { PropsWithRef } from "../../../types";
 import { UploadDraggerProps } from "./types";
 import usePropsWithThemeDefaults from "../../../motif/hooks/usePropsWithThemeDefaults";
+import { Validations } from "src/lib";
+import { mapExternalValue, toFormValue } from "@/components/Upload/helper.ts";
+import { FileType } from "@/components/Upload/types.ts";
 
 const UploadDragger = (p: PropsWithRef<UploadDraggerProps, HTMLDivElement>) => {
```
atakanbayrak, line 17: Does not make it hard to understand component work's mechanism when there are 3 files already uploaded and text message which tells there could be max 2 files available to upload ?  

Image

atakanbayrak: And I understand that, deletion of already uploaded files are not possible ? Logically that's so right but I am not sure users would like to have it anyway 🫨

src/lib/components/Upload/UploadInput/UploadInput.test.tsx
```diff
@@ -370,4 +372,49 @@ describe("UploadInput", () => {
 
     xhrSpy.mockRestore();
   });
+
+  it("should display the file name when a single file is given in the value prop", () => {
```
atakanbayrak, line 376: I am not sure how to do it and it is not the case of this issue but we could find a way to make these tests common for Upload components. Logic is same and we multiple each time with each other. Do you consider that we could open issue for it and discuss it advance ?

atakanbayrak (changes requested):
There was nothing to wrong with work mechanism's of components with this new feature, there are only few things that I would like to discuss in advance.

src/lib/components/Upload/UploadDragger/UploadDragger.tsx
aktasmehmet: Nope, this should be. There is a bug here I guess. If the component is not readonly or disabled, deletion should also be possible. Here the issue is probably a premature validation to block the component and make it readonly... Investigating... Good catch 😉

src/lib/components/Upload/helper.ts
aktasmehmet: Actually, FileType is very clear but I've put one for FileObject. Will be ready with the upcoming commit

src/lib/components/Upload/UploadDragger/UploadDragger.test.tsx
aktasmehmet: This is a basic delete scenario. I guess It has nothing to do with the files on the server. What do you say?

src/lib/components/Upload/UploadDragger/UploadDragger.tsx
aktasmehmet: Regarding with the 2 files 3 files thing, this is something that the user should aware of. Because, executing the validation for an already uploaded files will cause more damage then this (like showing one of the files as errored, etc.)

aktasmehmet: There arent any bugs there :) You've tested it with readonly prop. So, you cannot delete :)

src/lib/components/Upload/UploadInput/UploadInput.test.tsx
aktasmehmet: The scenarios are similar but the rendered items and their layout is different. I've thought about the same thing, in fact did a little bit development there, but stuck at some point. 

So, at this time, I prefer to keep as it is but maybe, we could think something else in the future. But I dont think we should create a ticket for that right now.

src/lib/components/Upload/UploadDragger/UploadDragger.tsx
atakanbayrak: That's true but when I try it without readonly right now, I didn't manage to delete a file anyway? Does deletion works good for your local ?

aktasmehmet: Yes it is fine. And I also tried to delete and it worked.

Screenshot 2026-06-22 at 12 19 51

atakanbayrak (approved):

ZehranurC (approved):

PR #90, opened 2026-06-18
fix: select responsive issue

ZehranurC:
## Description

- Resolved Select component responsiveness issue by fixing flex item sizing. The component now adapts correctly to container width without freezing at fixed dimensions.

- The `.inputText` flex item used the browser default `min-width: auto`, which prevented it from shrinking below content width.

- Added `width: 0;` to `.inputText` in Select.module.scss, allowing the flex item to shrink responsively while `flex-grow: 1;` handles expansion.

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [x] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

### Before

image

### After

image

## Checklist

## How to Test

 ```
<Form
            onSubmit={onSubmit}
            dontClearOnSubmit
            submitButtonLabel="Submit Application"
            buttonPosition="right"
            labelOrientation="horizontal"
            alternateButtons={[
              <Button
                key="1"
                label="Go Back"
                onClick={() => router.back()}
                shape="outline"
                variant="primary"
              />,
            ]}
          >
            <Grid fluid leanToEdge>
              <Grid.Row>
                <Grid.Col size={12} md={6}>
                  <Form.Field
                    label="Application Number"
                    name="basvuruNumarasi"
                    validations={[Validations.Required]}
                    helperText="Enter your application number"
                  >
                    <InputText placeholder="Enter number..." />
                  </Form.Field>
                </Grid.Col>
                <Grid.Col size={12} md={6}>
                  <Form.Field
                    label="Passport Number"
                    name="pasaportNumarasi"
                    validations={[Validations.Required]}
                    helperText="Enter your passport number"
                  >
                    <InputText placeholder="Enter passport number..." />
                  </Form.Field>
                </Grid.Col>
              </Grid.Row>

              <Grid.Row>
                <Grid.Col size={12} md={6}>
                  <Form.Field
                    label="Passport Country"
                    name="pasaportUlkesi"
                    validations={[Validations.Required]}
                    helperText="Select the country that issued your passport"
                  >
                    <Select 
                      data={countries} 
                      filterable 
                      placeholder="Select a country..." 
                    />
                  </Form.Field>
                </Grid.Col>
                <Grid.Col size={12} md={6}>
                  <Form.Field
                    label="Email Address"
                    name="epostaAdresi"
                    validations={[Validations.EMAIL, Validations.Required]}
                    helperText="Enter your email address"
                  >
                    <InputText 
                      type="email"
                      placeholder="example@email.com"
                      iconLeft="mail"
                    />
                  </Form.Field>
                </Grid.Col>
              </Grid.Row>
            </Grid>
          </Form>
```

### CODE REVIEW

File: src/lib/components/Select/Select.module.scss [hunk 1]
```diff
@@ -72,6 +72,7 @@ $sizes: (
 
   .inputText {
     all: unset;
+    width: 0;
```
aktasmehmet, line 75: This is not an elegant solution, besides it already has bugs. You've only focused on using it in a container but giving the width 0 will cause it to render only with the width of the paddings. Just check the screenshot. This is the usage:

```
<Select data={[{ label: "Türkiye", value: "TR" }]} filterable placeholder="Select a country..." />
```

Image

aktasmehmet (changes requested):

src/lib/components/Select/Select.module.scss
ZehranurC: Fixed standalone usage with new commit

aktasmehmet: Yes it is working but giving the width as 0 is not an optimum scenario.

Actually the issue here is that, inputs and selects (natively) has a minimum width in browsers (I guess it was 20). Since we put an additional dropdown icon to the right, select always has a wider minimum width than the inputs. 

This native minimum size blocks other flex elements. It overrides them. Also, another issue is that, in your solution, display becomes block, instead of inline-block. This may lead some issues since the initial value of native select is not block.

So, my suggestion is:

- Revert everything!
- Delete ``&.inFormField`` in scss completely. (We are now using native min-width of the elements. We are removing this class names from everywhere)
- Put a conditional size={1} to the Select itself to shrink. If I'm not wrong, everything will work as expected?

Like:

```
<input
            size={inFormField ? 1 : undefined}
```

aktasmehmet: Aaahh, this time horizontal form shrink will break :(

aktasmehmet: Anyway, your solution is the only fix here right now, approving 😉

aktasmehmet (approved):

PR #91, opened 2026-06-19
test(Stepper): component and useStepper tests are added

aktasmehmet:
## Description

Tests that cover ``Stepper`` component and ``useStepper`` companion has been added.

_NOTE: In order to keep things simple (since there are too many tests) I had to split the tests into 2 describe blocks._

## Type of Change

- [ ] 🆕 New Component
- [x] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Screenshots / Preview

Screenshot 2026-06-19 at 14 13 30
Screenshot 2026-06-19 at 14 13 12

## Checklist

## How to Test

Run jest

### DIFF
--- a/src/lib/components/Stepper/Stepper.test.tsx
+++ b/src/lib/components/Stepper/Stepper.test.tsx
@@ -87,15 +87,12 @@ describe("Stepper", () => {
   });
 
   it("should render step counter when showCount prop is true", () => {
-    const { container, rerender } = renderExt();
-    expect(container.querySelector(".stepCount")).not.toBeInTheDocument();
+    renderExt();
+    expect(screen.queryByText("/ 3")).not.toBeInTheDocument();
+    cleanup();
 
-    rerender(
-      <Stepper showCount>
-        <Stepper.Item title="Step 1">Content 1</Stepper.Item>
-      </Stepper>,
-    );
-    expect(container.querySelector(".stepCount")).toBeInTheDocument();
+    renderExt({ showCount: true });
+    expect(screen.getByText("/ 3")).toBeInTheDocument();
   });
 
   it("should render the correct initial active step using defaultActiveStep", () => {
@@ -104,6 +101,15 @@ describe("Stepper", () => {
     expect(stepItems[0]).toHaveClass("completed");
     expect(stepItems[1]).toHaveClass("active");
     expect(stepItems[2]).toHaveClass("upcoming");
+  });
+
+  it("should clamp defaultActiveStep to valid range", () => {
+    const { container, unmount } = renderExt({ defaultActiveStep: -1 });
+    expect(container.querySelectorAll(".stepItem")[0]).toHaveClass("active");
+    unmount();
+
+    const { container: highContainer } = renderExt({ defaultActiveStep: 99 });
+    expect(highContainer.querySelectorAll(".stepItem")[2]).toHaveClass("active");
   });
 
   it("should render step content of the active step only", () => {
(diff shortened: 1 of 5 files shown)


src/lib/components/Stepper/Stepper.test.tsx
```diff
@@ -0,0 +1,366 @@
...
+
+    // itemOrientation = vertical
+    expect(getRoot()).toHaveClass("vertical-items");
+
+    // variant = primary
+    expect(container.querySelector(".active")).toHaveClass("primary");
+  });
+
+  it("should render in the orientation given in orientation prop", () => {
+    const { rerender, getRoot } = renderExt({ orientation: "horizontal" });
```
atakanbayrak, line 50: Do we need to give horizontal prop specifically because it is already default value in here?

src/lib/components/Stepper/Stepper.test.tsx
```diff
@@ -0,0 +1,366 @@
...
+    (["primary", "secondary", "success", "warning", "danger"] as const).forEach(variant => {
+      const { container, unmount } = renderExt({ variant });
+      expect(container.querySelector(".active")).toHaveClass(variant);
+      unmount();
+    });
+  });
+
+  it("should render step counter when showCount prop is true", () => {
+    const { container, rerender } = renderExt();
+    expect(container.querySelector(".stepCount")).not.toBeInTheDocument();
```
atakanbayrak: querySelector based container selections have least hierarchy in test coverage after test id's. Is there any better way to track and check this situation?

src/lib/components/Stepper/Stepper.test.tsx
```diff
@@ -0,0 +1,366 @@
...
+
+    rerender(
+      <Stepper showCount>
+        <Stepper.Item title="Step 1">Content 1</Stepper.Item>
+      </Stepper>,
+    );
+    expect(container.querySelector(".stepCount")).toBeInTheDocument();
+  });
+
+  it("should render the correct initial active step using defaultActiveStep", () => {
```
atakanbayrak, line 98: defaultActiveStep prop is number based so we could add two more tests which would check if there is negative number or number which is more than step number.

src/lib/components/Stepper/Stepper.test.tsx
```diff
@@ -0,0 +1,366 @@
...
+    const { container } = renderExt({ defaultActiveStep: 1 });
+    const stepItems = container.querySelectorAll(".stepItem");
+    expect(stepItems[0]).toHaveClass("completed");
+    expect(stepItems[1]).toHaveClass("active");
+    expect(stepItems[2]).toHaveClass("upcoming");
+  });
+
+  it("should render step content of the active step only", () => {
```
atakanbayrak, line 115: We could enhance this test to check mid content because of previous and next content appearance. Previous step should be disappeared. Next step is already checked in here.

atakanbayrak (changes requested):

src/lib/components/Stepper/Stepper.test.tsx
aktasmehmet: yes necessary. since this is not a default value check, all props should be given. It is like checking size props of a button.

src/lib/components/Stepper/Stepper.test.tsx
aktasmehmet: Yes but a little complicated. Since there are 2 seperate spans there, it is not possible to check with a text. I am implementing to check the item with the right side of the text, after the slash.

src/lib/components/Stepper/Stepper.test.tsx
aktasmehmet: This is a topic of another test I think. This test should just check the content of the active step is rendered. The scenario you mentioned are covered in lots of other places. The first example comes to my mind is: "should navigate to previous step when the previous button is clicked". and there are more similar

src/lib/components/Stepper/Stepper.test.tsx
aktasmehmet: It was not checked before, thats the reason of the missing test. I'm implementing it. Upper bound in useStepper will not be able to be clamped because of the nature of that hook. It doesnt know the number of items. But in the component itself, it will work as expected.

atakanbayrak (approved):

PR #97, opened 2026-07-02
fix(DateRangePicker): implement responsive design

ZehranurC:
## Description

- Implement Figma's responsive layout for desktop and mobile views.
- Update tests to cover responsive behavior.

## Type of Change

- [ ] 🆕 New Component
- [x] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

### Before
image

### After
image

## Checklist


src/lib/components/DateRangePicker/DateRangePicker.module.scss
```diff
@@ -130,3 +130,68 @@ $sizes: (
...
+
+      > button[class*="arrowButton"]:first-of-type {
+        order: 2;
+      }
+
+      > button[class*="arrowButton"]:last-of-type {
+        order: 3;
+      }
+
+      > div[class*="headerMonthYear"] {
```
atakanbayrak, line 191: It will be string based control mechanism in the future for css based situation. I think there should be better way to keep them on track instead of className naming.

src/lib/components/DateRangePicker/DateRangePicker.test.tsx
```diff
@@ -364,4 +364,51 @@ describe("DateRangePicker", () => {
...
+
+  it("should render correctly on desktop view (>768px) with date inputs", () => {
+    Object.defineProperty(window, "innerWidth", { writable: true, value: 1024 });
+    const mockStartDate = new Date(year, month, 5);
+    const mockEndDate = new Date(year, month, 15);
+    const { getFirstInput, getSecondInput } = renderExt(<DateRangePicker value={[mockStartDate, mockEndDate]} />);
+
+    expect(getFirstInput()).toBeInTheDocument();
+    expect(getSecondInput()).toBeInTheDocument();
+  });
```
atakanbayrak, line 388: These could be on the same tests without make them double.

src/lib/components/Motif/Pickers/Picker.module.scss
```diff
@@ -22,6 +24,13 @@ $sizes-picker: (("xs", "4x"), ("sm", "8x"), ("md", "12x"), ("lg", "16x"));
       padding-top: map.get(pickerValues.$size, $name, "gapV");
       gap: map.get(pickerValues.$size, $name, "gapV");
```
atakanbayrak, line 25: We could also define const for these values.

atakanbayrak (changes requested):

src/lib/components/DateRangePicker/DateRangePicker.tsx
```diff
@@ -21,10 +22,13 @@ const DateRangePicker = (props: PropsWithRef<DateRangePickerProps, HTMLDivElemen
     style,
   } = usePropsWithThemeDefaults("DateRangePicker", props);
 
+  const screenSize = useScreenSize();
+  const isCompactView = screenSize.width > 0 ? screenSize.width <= 768 : false;
```
emreekiziltoprak, line 26: a hook can be used for size changes and querying media dynamically in javascript side with browsers matchmedia, also this hook should fix the initial render problem
```
 
"use client";
import { useState, useEffect } from "react";

const useMediaQuery = (query: string) => {
  const [matches, setMatches] = useState(false);

  useEffect(() => {
    if (typeof window === "undefined") return;

    const mediaQueryList = window.matchMedia(query);
    setMatches(mediaQueryList.matches);

    const listener = (event: MediaQueryListEvent) => {
      setMatches(event.matches);
    };

    mediaQueryList.addEventListener("change", listener);

    return () => {
      mediaQueryList.removeEventListener("change", listener);
    };
  }, [query]);

  return matches;
};

export default useMediaQuery;

const isCompactView = useMediaQuery("(max-width: 768px)");
```

src/lib/components/Motif/Pickers/Picker.module.scss
```diff
@@ -7,13 +7,15 @@ $sizes-picker: (("xs", "4x"), ("sm", "8x"), ("md", "12x"), ("lg", "16x"));
 
 @each $name, $padding in $sizes-picker {
```
emreekiziltoprak, line 8: using scss mapping feature for only adjusting variables decreases bundles size. doing style assignment with mapping in main class can cause longer runtime. making style adjustment at once should improve performance as a suggestion

src/lib/components/DateRangePicker/DateRangePicker.tsx
```diff
@@ -21,10 +22,13 @@ const DateRangePicker = (props: PropsWithRef<DateRangePickerProps, HTMLDivElemen
     style,
   } = usePropsWithThemeDefaults("DateRangePicker", props);
 
+  const screenSize = useScreenSize();
```
aktasmehmet, line 25: When I see this, I immediately stop reviewing and look for a pure css solution. And did it. Because of that, I stopped review and want to discuss with you this solution. Most important difference is that I used pure css, keep 2 additional picker as it is. Because user may play with the width and burden on the performance side is bigger if we watch all the width changes. I've put all the changes I've made below.

Please note that I did not think about the sizes, did not think about the px usage... It is just for a small MVP. Here is the ss of the outcome:

Image

**DateRangePicker.module.scss**
```
@include screen-width("below", "md") {
  .selectionBar {
    text-align: initial;
    display: flex;
  }
  .rangeSelectDropdown { display: none !important; }
  .datePickers {
    gap: 24px;
    transform: translateX(calc(-100% - 24px)) !important;
    > div:nth-child(2) > div:first-of-type > button:last-of-type {
      display: block !important;
    }
    &.slideLeft { transform: translateX(0) !important; }
    &.slideRight { transform: translateX(0) !important; }
  }
}
```

**CustomDatePicker.tsx**
```
onPrevClick={onPrevClick}
onNextClick={onNextClick}
```

**DaysDropdown.tsx**
```
return <Dropdown size={size} items={items} variant="secondary" label={label} className={styles.rangeSelectDropdown} />;
```

**Picker.module.scss**
```
@include screen-width("below", "md") {
  .wide {
    max-width: 342px !important;
    min-width: 342px !important;
  }
}
```

There were also some bugs on your side, like when using next arrow, global css usage, rerenders, etc. Here are the recording and ss of it

https://github.com/user-attachments/assets/c3a0e56c-0df3-452c-baed-2e47e0c04670
Image

aktasmehmet (changes requested):

PR #100, opened 2026-07-13
fix(InputText): clear button icon click area overflow fix

ZehranurC:
## Description

- Fixed clear button icon overflow in InputText using CSS custom property --clear-icon. 
- Issue found in all input components using InputText (InputDate, InputPassword, etc.), not just InputDate.

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [x] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Checklist

## How to test
Test with Layout component using material icons. Use the form example below in page.tsx to verify the fix works across different input types with clearable buttons.

```
import InputText from "../lib/components/InputText";
import InputDate from "../lib/components/InputDate";
import InputPassword from "../lib/components/InputPassword";
import Button from "../lib/components/Button";
import Form from "../lib/components/Form";

export default function ClearButtonTest() {
  const handleSubmit = (data: any) => {
    console.log("Submitted:", data);
  };

  return (
    <div style={{ maxWidth: "500px", margin: "0 auto", padding: "20px" }}>
      <Form onSubmit={handleSubmit}>
        <Form.Field name="text" label="Text">
          <InputText name="text" placeholder="Enter text" clearable />
        </Form.Field>

        <Form.Field name="date" label="Date">
          <InputDate name="date" />
        </Form.Field>

        <Form.Field name="password" label="Password">
          <InputPassword name="password" clearable />
        </Form.Field>

        <Button label="Submit" htmlType="submit" />
      </Form>
    </div>
  );
}

```


emreekiziltoprak (dismissed):

atakanbayrak (dismissed):
After chromatic good to go

src/lib/components/Icon/Icon.module.scss
```diff
@@ -4,7 +4,7 @@ $variants: "primary", "secondary", "info", "success", "warning", "danger";
 $sizes: (("xxs", "16x"), ("xs", "18x"), ("sm", "20x"), ("md", "22x"), ("lg", "24x"), ("xl", "28x"), ("xxl", "32x"));
 
 .Root {
-  display: inline-flex !important;
+  display: var(--clear-icon, inline-flex);
```
aktasmehmet: This has nothing to do with the icon component itself. The positioning issue is a problem of its parent. One thing to consider regardsless of this task. Display should be "inline", not "inline-flex" because it is a basic span.

src/lib/components/Motif/InputText/InputText.module.scss
```diff
@@ -114,6 +114,13 @@ $sizes: (("xs", "8x", "4x", "18x"), ("sm", "12x", "6x", "20x"), ("md", "16x", "8
...
   position: absolute;
   right: 0;
+  --clear-icon: block;
+
+  span {
+    overflow: hidden;
+    line-height: 1;
+    text-align: center;
+  }
 }
```
aktasmehmet, line 119: I would clear everything you've put here. Just adding the code below should solve the issue:
```
display: inline-flex;
  align-self: center;
```

aktasmehmet (changes requested):

src/lib/components/Icon/Icon.module.scss
aktasmehmet: I just saw the comment of Claude and it is right. We should not take the risk of affecting other important components at this time. So, reverting this line back to its original would be a better solution.

src/lib/components/Motif/InputText/InputText.module.scss
```diff
@@ -114,6 +114,11 @@ $sizes: (("xs", "8x", "4x", "18x"), ("sm", "12x", "6x", "20x"), ("md", "16x", "8
 .clearButton {
   position: absolute;
   right: 0;
+
+  span {
+    display: inline-flex;
+    align-self: center;
```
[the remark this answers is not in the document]
aktasmehmet: After the last merge on main, this part is fixed. So, please:

1 - Remove this span section
2 - Test again

aktasmehmet (changes requested):
waiting for open threads

src/lib/components/Icon/Icon.module.scss
```diff
@@ -4,7 +4,7 @@ $variants: "primary", "secondary", "info", "success", "warning", "danger";
 $sizes: (("xxs", "16x"), ("xs", "18x"), ("sm", "20x"), ("md", "22x"), ("lg", "24x"), ("xl", "28x"), ("xxl", "32x"));
 
 .Root {
-  display: inline-flex !important;
+  display: inline-flex;
```
aktasmehmet: Why dont you revert this completely and leave a change each time? This need a full revert.

aktasmehmet (changes requested):

atakanbayrak (approved):

aktasmehmet (approved):

PR #103, opened 2026-07-14
fix(PinCode): correct sizing calculations and match Figma designs

ZehranurC:
## Description

Fixed PinCode component sizing issues and aligned calculations with Figma specifications. All sizes (xs, sm, md, lg) now display correctly at their intended pixel values (28px, 32px, 38px, 44px).

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [x] 🧹 Code Refactoring

### Before 
image

### After
image

## Checklist

```
"use client";

import { useMotifContext } from "../lib/motif/context/MotifProvider";
import { PinCode } from "../lib";
import { useState } from "react";

const Home = () => {
  const { t } = useMotifContext();
  const [pinCodeValue, setPinCodeValue] = useState<string[]>(["", "", "", "", "", ""]);

  return (
    <div style={{ margin: "0 auto", padding: 20 }}>
      <h2>{t("g.hello_x", { name: "MOTİF UI" })}</h2>
      <h4>{t("misc.playgroundDescription")}</h4>

      <div style={{ marginTop: 40 }}>
        <h3>PinCode Size Examples</h3>

        <div style={{ marginBottom: 30 }}>
          <h4>XS (28px)</h4>
          <PinCode size="xs" value={pinCodeValue} onChange={val => setPinCodeValue(Array.isArray(val) ? val : [])}>
            <PinCode.Item />
            <PinCode.Item />
            <PinCode.Item />
            <PinCode.Item space />
            <PinCode.Item />
            <PinCode.Item />
          </PinCode>
        </div>

        <div style={{ marginBottom: 30 }}>
          <h4>SM (32px)</h4>
          <PinCode size="sm" value={pinCodeValue} onChange={val => setPinCodeValue(Array.isArray(val) ? val : [])}>
            <PinCode.Item />
            <PinCode.Item />
            <PinCode.Item />
            <PinCode.Item space />
            <PinCode.Item />
            <PinCode.Item />
          </PinCode>
        </div>

        <div style={{ marginBottom: 30 }}>
          <h4>MD (38px - Default)</h4>
          <PinCode value={pinCodeValue} onChange={val => setPinCodeValue(Array.isArray(val) ? val : [])}>
            <PinCode.Item />
            <PinCode.Item />
            <PinCode.Item />
            <PinCode.Item space />
            <PinCode.Item />
            <PinCode.Item />
          </PinCode>
        </div>

        <div style={{ marginBottom: 30 }}>
          <h4>LG (44px)</h4>
          <PinCode size="lg" value={pinCodeValue} onChange={val => setPinCodeValue(Array.isArray(val) ? val : [])}>
            <PinCode.Item />
            <PinCode.Item />
            <PinCode.Item />
            <PinCode.Item space />
            <PinCode.Item />
            <PinCode.Item />
          </PinCode>
        </div>
      </div>
    </div>
  );
};

export default Home;

```


src/lib/components/PinCode/PinCode.module.scss
```diff
@@ -13,9 +13,9 @@ $sizes: (
     gap: var(--base-sizing-#{$gap});
     .input input {
       @include font("body", $size);
-      width: calc(var(--base-sizing-#{$inputSize}) + 2 * var(--base-sizing-#{$padding}) + var(--base-sizing-2x));
-      height: calc(var(--base-sizing-#{$inputSize}) + 2 * var(--base-sizing-#{$padding}) + var(--base-sizing-2x));
-      padding: var(--base-sizing-#{$padding});
+      width: calc(var(--base-sizing-#{$inputSize}) + 2 * var(--base-sizing-#{$padding}) - var(--base-sizing-2x));
```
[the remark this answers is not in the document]
atakanbayrak: It looks like Claude says correct but do we sure that component reflects sizes which are given inside PR description. I mean, Claude calculations seems legit. Better to check one more time.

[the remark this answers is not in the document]
atakanbayrak: Moreover we deleted padding to have a solution with only width. We have to be sure padding 0 does not affect anything else.

atakanbayrak (changes requested):

emreekiziltoprak:
resetting padding as 0 may be is not necessary

src/lib/components/PinCode/PinCode.module.scss
```diff
@@ -11,19 +11,26 @@ $sizes: (
...
     .input input {
       @include font("body", $size);
-      width: calc(var(--base-sizing-#{$inputSize}) + 2 * var(--base-sizing-#{$padding}) + var(--base-sizing-2x));
-      height: calc(var(--base-sizing-#{$inputSize}) + 2 * var(--base-sizing-#{$padding}) + var(--base-sizing-2x));
-      padding: var(--base-sizing-#{$padding});
+      width: var(--base-sizing-#{$inputSize});
+      height: var(--base-sizing-#{$inputSize});
+      padding: 0;
+      flex: none;
     }
```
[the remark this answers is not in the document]
atakanbayrak: In practice there will be no change when we compare with the solution appears now but Claude refers something possibly happen. We could think about subtracting the border explicitly.

atakanbayrak (changes requested):
Subtracting the border should be considered, after ready to go.

src/lib/components/PinCode/PinCode.module.scss
```diff
@@ -11,19 +11,26 @@ $sizes: (
...
   }
 }
 
 .Root {
   display: inline-flex;
   .input input {
-    box-sizing: content-box;
+    box-sizing: border-box;
```
aktasmehmet, line 31: This waw causing some issues when using in a real project. I cannot remember the exact issue but, @atakanbayrak do you recall? I'll continue revieweing depending on your answer.

aktasmehmet (changes requested):

src/lib/components/PinCode/PinCode.module.scss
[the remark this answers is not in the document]
ZehranurC: Math inconsistency var ama pratikte doğru (spec match).
-2px tested: failed.

src/lib/components/PinCode/PinCode.module.scss
atakanbayrak: If you mean using border-box instead of content-box, there was box-sizing usage inside default-theme.css which causes error in layer structure, but this is not the case in here. I don't remember anything else about this box-sizing issue. I checked the tags of PinCode file and didn't find out anything else.

aktasmehmet: alright, I am resolving here, then. thx

src/lib/components/PinCode/PinCode.module.scss
```diff
@@ -11,19 +11,26 @@ $sizes: (
 @each $size, $padding, $inputSize, $gap in $sizes {
   .#{$size} {
     gap: var(--base-sizing-#{$gap});
+    .input {
+      width: calc(var(--base-sizing-#{$inputSize}) + 2 * var(--base-sizing-#{$padding}));
+      height: calc(var(--base-sizing-#{$inputSize}) + 2 * var(--base-sizing-#{$padding}));
+      padding: var(--base-sizing-#{$padding});
+      box-sizing: border-box;
+    }
```
aktasmehmet: InputWrappers are smaller than the input itself in md and xs. Actually we have to figure this out completely. Instead of giving the fixed width here, we can only use the width and height you just gave below, in ".input input" part. Padding will handle the rest. So you can remove the width and height here. There will be 2 pixel difference with figma because the of the border difference

Image
Image

aktasmehmet (changes requested):

src/lib/components/PinCode/PinCode.module.scss
ZehranurC: I had given up on applying the solution you mentioned because of the border difference, but since we're on the same page, I'm fixing it this way

aktasmehmet (approved):

atakanbayrak (approved):

PR #104, opened 2026-07-17
fix(NavBarSearch): clearable button is added

emreekiziltoprak:
## Description

Although there is input text in the NavBar search field, the clearable feature is not supported.

## Type of Change

- [ ] 🆕 New Component
- [x] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Checklist


src/lib/components/NavBar/types.ts
atakanbayrak, line 25: While I was checking if we could determine clearable prop to default, it is not possible to make it happen because we define different type for search props inside NavBar. It might be needed to use of clearable prop by default in the future so what could be the best approach to make either some of the search props defaultable? More opinion needed, @motif-ui/core-team 

1 - We could create different default able type for Search props which will be overwhelming at the future.
2- We could use clearable prop inside default able porps and phrase it to default able props with omitting other props.

...

src/lib/components/NavBar/NavBar.test.tsx
```diff
@@ -153,6 +153,31 @@ describe("NavBar", () => {
...
+    expect(queryByText("cancel_outline")).not.toBeInTheDocument();
+  });
+
+  it("should render a clear button in the search input when clearable is true in the search prop and text has been typed", async () => {
+    const { getByRole, queryByText } = render(<NavBar search={{ clearable: true }} />);
+    const searchInput = getByRole("textbox") as HTMLInputElement;
+    expect(queryByText("cancel_outline")).not.toBeInTheDocument();
+    await userEvent.type(searchInput, "test query");
+    expect(queryByText("cancel_outline")).toBeInTheDocument();
+  });
```
atakanbayrak: We could check if there is no clear button and clear button together inside the same test.

1- We could check clearable button when there is no text.
2- We could check clearable button when text is typed.

atakanbayrak (changes requested):
Someone finally used to check component from role, thank you. 😄

src/lib/components/NavBar/components/NavBarSearch.tsx
```diff
@@ -13,6 +13,7 @@ const NavBarSearch = (props: NavBarSearchProps) => {
     placeholder = t("g.search"),
     onPressEnter,
     pill,
+    clearable,
```
atakanbayrak: Clearable prop does not have to pass through with props because it will be implemented implicitly. This line should be reverted.

src/lib/components/NavBar/components/NavBarSearch.tsx
```diff
@@ -49,8 +50,10 @@ const NavBarSearch = (props: NavBarSearchProps) => {
       <InputText
         size="md"
         pill={pill}
+        clearable={clearable}
```
atakanbayrak:
```suggestion
        clearable
```

src/lib/components/NavBar/NavBar.stories.tsx
```diff
@@ -23,6 +23,7 @@ const argValues: NavBarProps = {
     onButtonClick: () => alert("Check 'NavBar/Search' menu on the left for more detail!"),
```
atakanbayrak, line 23: All changes in this file should be reverted.

1- Clearable prop does not have to pass through with props because it is already driller inside InputText as I mentioned.
2- Clearable prop will not be shown to end user to manipulate it so it is not needed inside stories props because of the reason as I reflected before.

src/lib/components/NavBar/NavBar.stories.tsx
```diff
@@ -114,13 +115,15 @@ return (
...
+      clearable: true,
     }}
   />
-);      
+);
   `,
       },
     },
   },
+
```
atakanbayrak, line 125: This is not related with the this PR but for the best reading and reviewing practice we try to check changes before we sent commit to delete empty lines and etc to increase reading and reviewing possibility. 

Moreover, I am not sure but do you have ESLint auto detection installed and enabled in your ide ? Because some errors has been shown by the ide to me and you cannot unseen them when you have ESLint auto detection enabled. If you don't have it, do not hesitate to contact with the team members it has kinda weird enable mechanism inside ide sometimes.

atakanbayrak (changes requested):
There are several issues I would like to talk about after check localhost.

src/lib/components/NavBar/components/NavBarSearch.tsx
ZehranurC: +1

src/lib/components/NavBar/NavBar.test.tsx
ZehranurC: +1

src/lib/components/NavBar/types.ts
ZehranurC: I think Option 2. Add clearable to NavBarSearchProps as optional and omit other unnecessary props. NavBarSearch being a self contained component eliminates duplication risk.

src/lib/components/NavBar/NavBar.stories.tsx
ZehranurC: +1

src/lib/components/NavBar/components/NavBarSearch.tsx
```diff
@@ -13,6 +13,7 @@ const NavBarSearch = (props: NavBarSearchProps) => {
     placeholder = t("g.search"),
```
aktasmehmet, line 13: @emreekiziltoprak regardless of the review, can you please fill the PR appropriately? (Description, before-after screenshots, etc.)

And rename the title like: 

```fix(NavBarSearch): clearable feature is added search box```

File: src/lib/components/NavBar/components/NavBarSearch.tsx (hunk 2 above)
Comment (aktasmehmet) [reply src/lib/components/NavBar/components/NavBarSearch.tsx]: +1

File: src/lib/components/NavBar/components/NavBarSearch.tsx (hunk 3 above)
Comment (aktasmehmet) [reply src/lib/components/NavBar/components/NavBarSearch.tsx]: +1

File: src/lib/components/NavBar/types.ts [hunk 7]
```diff
@@ -99,6 +100,7 @@ export type NavBarMenuContextProps = {
 export type NavBarSearchProps = {
   placeholder?: string;
   pill?: boolean;
+  clearable?: boolean;
```
aktasmehmet: At this moment, I dont think this should be optional. We should directly put it into the component. So, this is not needed.

src/lib/components/NavBar/types.ts
aktasmehmet: I think we force that, remove this prop entirely and make the input clearable, thats all.

src/lib/components/NavBar/components/NavBarSearch.tsx
```diff
@@ -49,8 +50,10 @@ const NavBarSearch = (props: NavBarSearchProps) => {
       <InputText
         size="md"
         pill={pill}
+        clearable={clearable}
         placeholder={placeholder}
         onChange={val => setSearchQuery(val as string)}
+        onClearClick={() => setMaybeShowResults(false)}
```
aktasmehmet: This is not the right approach. You only hide the previous results with this. ``searchQuery`` state stays the same. This is working by chance because the search results only open when there is another text in the input.

So, you should call ``setSearchQuery("")`` here.

src/lib/components/NavBar/NavBar.test.tsx
aktasmehmet: Agree, you may do that 😉

aktasmehmet (changes requested):

src/lib/components/NavBar/types.ts
atakanbayrak: That's an another option we could think of if we are agree to force it. If you ask me it doesn't bother me to do it that way.

src/lib/components/NavBar/components/NavBarSearch.tsx
```diff
@@ -49,8 +49,13 @@ const NavBarSearch = (props: NavBarSearchProps) => {
       <InputText
         size="md"
         pill={pill}
+        clearable
         placeholder={placeholder}
         onChange={val => setSearchQuery(val as string)}
+        onClearClick={() => {
+          setSearchQuery("");
+          setMaybeShowResults(false);
```
aktasmehmet: This is a little bit tricky. Use scenario of the search is like:

"User feeds the result after something is triggered by Navbar (like pressenter or press search button)"

So, user should decide what to do with the clear process. We should only clear the search query here. And introduce a new prop "onClear" and fire that event when this is clicked. Then user will decise what to do with the results and the "showing result".

Of course you should change the tests accordingly.

aktasmehmet (changes requested):

src/lib/components/NavBar/components/NavBarSearch.tsx
```diff
@@ -16,6 +16,7 @@ const NavBarSearch = (props: NavBarSearchProps) => {
     results = [],
     onResultClick,
     onButtonClick,
+    onClear,
```
aktasmehmet, line 19: Perfect! One final request. We have a search story for Navbar spesifically to show the search functionality. It would be great to implement onClear callback and clear the results there. Then we are good to go

atakanbayrak (dismissed):

aktasmehmet (approved):
Good work 👏

PR #105, opened 2026-07-17
test(Form, Upload): tests are added for value props

emreekiziltoprak:
## Description

Relevant tests must be written or updated for any scenario where a file can be set via the `value` prop in upload components (including `UploadList`, `UploadInput`, `UploadDragger`, and instances where they are used within a `Form`). Additionally, sections concerning the `disabled` and `readonly` props should be reviewed and updated if necessary.

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Checklist


src/lib/components/Form/test/Form.test.tsx
```diff
@@ -1410,6 +1411,107 @@ describe("Form", () => {
...
+  });
+
+  it("should include both value files and newly added files in submit data for UploadList and UploadDragger", async () => {
+    const xhrSpy = mockXHRs();
+    const serverFile = { id: "server-1", name: "server-doc.pdf", type: "application/pdf", size: 2048 };
+    const handleSubmit = jest.fn((data: FormSubmitData) => {
+      /* data.values = {
+      uploadList:    [ serverFile, test.pdf ],   
+      uploadDragger: [ serverFile, test.pdf ],   
+    } */
```
atakanbayrak: Comment line is forgotten.

src/lib/components/Form/test/Form.test.tsx
```diff
@@ -1410,6 +1411,107 @@ describe("Form", () => {
...
+      </Form>,
+    );
+
+    //files are should be current
+    expect(screen.queryAllByText(serverFile.name)).toHaveLength(3);
+    //when field is disabled delete button is not shown
+    expect(screen.queryAllByText("delete")).toHaveLength(0);
+
+    await userEvent.setup().click(screen.getByText(t("g.submit")));
+  });
```
atakanbayrak: We could merge this test with readOnly prop and find a naming reflects disabled and readOnly prop because their attitude is the same.

src/lib/components/Upload/UploadDragger/UploadDragger.test.tsx
```diff
@@ -424,4 +424,41 @@ describe("UploadDragger", () => {
     await simulateDrop(getDragArea(), [MOCK.filePdf1kb]);
     expect(getFileList()?.childNodes).toHaveLength(1);
   });
+
+  it("should show both the value file and a newly dropped file together and upload only the new one", async () => {
```
atakanbayrak:
```suggestion
  it("should display the existing value file alongside the newly dropped file, but only upload the new one", async () => {
```

src/lib/components/Upload/UploadDragger/UploadDragger.test.tsx
```diff
@@ -424,4 +424,41 @@ describe("UploadDragger", () => {
...
+    await simulateDrop(getDragArea(), [MOCK.filePdf1kb]);
+    expect(getFileList()?.childNodes).toHaveLength(2);
+    await waitForSuccessfulUpload(getFileItemLast());
+
+    expect(getFileItemFirst()).toHaveTextContent(serverFile.name);
+    expect(getFileItemFirst()).toHaveTextContent(formatBytes(serverFile.size));
+    expect(getFileItemFirst()).not.toHaveTextContent(t(MESSAGE.UPLOAD_SUCCESS));
+  });
+
+  it("should remove the disabled state of the drag area when the value file filling the maxFile limit is deleted", async () => {
```
atakanbayrak:
```suggestion
  it("should re-enable the drag area after deleting the value file that was filling the maxFile limit", async () => {
```

src/lib/components/Upload/UploadDragger/UploadDragger.test.tsx
```diff
@@ -424,4 +424,41 @@ describe("UploadDragger", () => {
...
+    expect(getDragArea()).toHaveClass("disabled");
+
+    await userEvent.click(getDeleteButton());
+    await waitFor(() => {
+      expect(getFileList()).not.toBeInTheDocument();
+      expect(getDragArea()).not.toHaveClass("disabled");
+    });
+  });
+
+  //value stays consistent ?
```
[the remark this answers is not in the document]
atakanbayrak: +1

src/lib/components/Upload/UploadDragger/UploadDragger.test.tsx
```diff
@@ -424,4 +424,41 @@ describe("UploadDragger", () => {
...
+  it("should not reflect value prop changes after mount", () => {
+    const { rerender, getFileList } = renderExt(<UploadDragger {...requiredProps} value={[serverFile]} maxFile={2} />);
+    expect(getFileList()?.childNodes).toHaveLength(1);
+
+    rerender(<UploadDragger {...requiredProps} value={[serverFile, serverFile2]} maxFile={2} />);
+    expect(getFileList()?.childNodes).toHaveLength(1);
+    expect(getFileList()).toHaveTextContent(serverFile.name);
+    expect(getFileList()).not.toHaveTextContent(serverFile2.name);
+  });
```
atakanbayrak: Test covers this statement but, shouldn't be reflect the value prop changes when there is value changes even component is mount? @motif-ui/core-team

src/lib/components/Upload/UploadInput/UploadInput.test.tsx
```diff
@@ -388,6 +388,18 @@ describe("UploadInput", () => {
...
+    const { getFileItem, getDeleteButton } = renderExt(<UploadInput {...requiredProps} value={[serverFile]} disabled />);
+    expect(getFileItem()).toHaveTextContent(serverFile.name);
+    expect(getDeleteButton()).not.toBeInTheDocument();
+  });
+
+  it("should hide the delete button but keep showing the file name when a value file is present and readOnly is true", () => {
+    const { getFileItem, getDeleteButton } = renderExt(<UploadInput {...requiredProps} value={[serverFile]} readOnly />);
+    expect(getFileItem()).toHaveTextContent(serverFile.name);
+    expect(getDeleteButton()).not.toBeInTheDocument();
+  });
```
atakanbayrak: I am not sure that we need these tests, because disabled does not affect any of the visibility issue except click and color.

atakanbayrak (changes requested):

src/lib/components/Upload/UploadList/UploadList.test.tsx
```diff
@@ -470,4 +471,40 @@ describe("UploadList", () => {
     await simulateChooseFiles(getInput(), [MOCK.filePdf1kb]);
     expect(getFileList()?.childNodes).toHaveLength(1);
   });
+
+  it("should show both the value file and a newly added file together and upload only the new one", async () => {
```
aktasmehmet:
```suggestion
  it("should upload only the new file when value prop is not undefined and a new file is browsed and added", async () => {
```

src/lib/components/Upload/UploadList/UploadList.test.tsx
```diff
@@ -470,4 +471,40 @@ describe("UploadList", () => {
...
+    );
+    expect(getFileList()?.childNodes).toHaveLength(1);
+
+    await simulateChooseFiles(getInput(), [MOCK.filePdf1kb]);
+    expect(getFileList()?.childNodes).toHaveLength(2);
+    await waitForSuccessfulUpload(getFileItemLast());
+
+    expect(getFileItemFirst()).toHaveTextContent(serverFile.name);
+    expect(getFileItemFirst()).toHaveTextContent(formatBytes(serverFile.size));
+    expect(getFileItemFirst()).not.toHaveTextContent(t(MESSAGE.UPLOAD_SUCCESS));
```
aktasmehmet, line 502: You may think to add some assertions about the newly uploaded file here. If the second item has success, etc.

src/lib/components/Upload/UploadList/UploadList.test.tsx
```diff
@@ -470,4 +471,40 @@ describe("UploadList", () => {
...
+    expect(getBrowseButton()).toBeDisabled();
+
+    await userEvent.click(getDeleteButton());
+    await waitFor(() => {
+      expect(getFileList()).not.toBeInTheDocument();
+      expect(getBrowseButton()).not.toBeDisabled();
+    });
+  });
+
+  it("should not reflect value prop changes after mount", () => {
```
aktasmehmet: Perfect and definitely forgotten scenario 👏 

But the thing is, we need to implement the opposite here. It is up to you to make that fix here or in a new task. If it is more complicated than it seems, you can create a ticket and put the id of it here. OR, you can fix this.

The thing is that all of our input components are reflecting the value changes. So, I personally thing that it should be the same here.

src/lib/components/Upload/UploadInput/UploadInput.test.tsx
aktasmehmet: These tests should definitely exist. But maybe we should move them elsewhere. We already have the tests below:

```should be rendered as disabled when disabled prop is true```
```should be rendered as readOnly when readOnly prop is true```

These tests (ideally) should cover all disabled and readonly scenarios. So I think that would be better to append a rerender and then use the codes here in those tests.

File: src/lib/components/Upload/UploadInput/UploadInput.test.tsx [hunk 11]
```diff
@@ -417,4 +429,18 @@ describe("UploadInput", () => {
...
+  it("should not reflect value prop changes after mount", () => {
+    const { rerender, getFileItem } = renderExt(<UploadInput {...requiredProps} value={[serverFile]} />);
+    expect(getFileItem()).toHaveTextContent(serverFile.name);
+
+    rerender(<UploadInput {...requiredProps} value={[serverFile2]} />);
+    expect(getFileItem()).toHaveTextContent(serverFile.name);
+    expect(getFileItem()).not.toHaveTextContent(serverFile2.name);
+  });
+
+  it("should not render a download action even when onDownloadClick is provided in value", () => {
```
aktasmehmet: This is also a perfect test! But this should not behave like this :( We need to implement a download button here, somehow. I've already created a task for this one: ```EDKUI-1479```. Putting this test (in a negative way) is also in that task.

So, we need to delete this test for now 😉

src/lib/components/Upload/UploadInput/UploadInput.test.tsx
```diff
@@ -417,4 +429,18 @@ describe("UploadInput", () => {
     await waitFor(() => expect(getDeleteButton()).not.toBeInTheDocument());
     xhrSpy.mockRestore();
   });
+
+  it("should not reflect value prop changes after mount", () => {
```
aktasmehmet: I've put the related comment to UploadList section. Same applies here, too.

src/lib/components/Upload/UploadDragger/UploadDragger.test.tsx
aktasmehmet: This test actually checks if only the new file is uploaded. But the naming makes me think different. We should either split this into 2 ("alongside view", "not upload the already given file") or make it like:

```should upload only the new file when value prop is not undefined and a new file is browsed and added```

File: src/lib/components/Upload/UploadDragger/UploadDragger.test.tsx [hunk 13]
```diff
@@ -424,4 +424,41 @@ describe("UploadDragger", () => {
...
+    );
+    expect(getFileList()?.childNodes).toHaveLength(1);
+
+    await simulateDrop(getDragArea(), [MOCK.filePdf1kb]);
+    expect(getFileList()?.childNodes).toHaveLength(2);
+    await waitForSuccessfulUpload(getFileItemLast());
+
+    expect(getFileItemFirst()).toHaveTextContent(serverFile.name);
+    expect(getFileItemFirst()).toHaveTextContent(formatBytes(serverFile.size));
+    expect(getFileItemFirst()).not.toHaveTextContent(t(MESSAGE.UPLOAD_SUCCESS));
```
aktasmehmet, line 443: You may think to add some assertions about the newly uploaded file here. If the second item has success, etc.

src/lib/components/Upload/UploadDragger/UploadDragger.test.tsx
aktasmehmet: I've put the same comment into UploadList side. I think so, we should reflect the value changes.

src/lib/components/Form/test/Form.test.tsx
```diff
@@ -1410,6 +1411,107 @@ describe("Form", () => {
     expect(getFormField(2)).not.toHaveClass("error");
   });
 
+  it("should keep value files visible but hide their delete actions when UploadInput, UploadList, and UploadDragger fields are disabled", async () => {
```
aktasmehmet: This is something component related I know but in order to make it more readible, I think it should better to keep all the disabled form item checks in the same test, here:

```
should disable all items in FormFieldGroup when disabled prop is given to FormFieldGroup
```

So, you should move these checks there with a comment of "Upload Components"

src/lib/components/Form/test/Form.test.tsx
```diff
@@ -1410,6 +1411,107 @@ describe("Form", () => {
...
+
+    //files are should be current
+    expect(screen.queryAllByText(serverFile.name)).toHaveLength(3);
+    //when field is disabled delete button is not shown
+    expect(screen.queryAllByText("delete")).toHaveLength(0);
+
+    await userEvent.setup().click(screen.getByText(t("g.submit")));
+  });
+
+  it("should keep value files visible but hide their delete actions when UploadInput, UploadList, and UploadDragger fields are readOnly", async () => {
```
aktasmehmet: same what I said in disabled test. There is also a readonly test to cover all the form items. You can move this into that.

src/lib/components/Form/test/Form.test.tsx
```diff
@@ -1410,6 +1411,107 @@ describe("Form", () => {
...
+      </Form>,
+    );
+
+    expect(screen.queryAllByText(serverFile.name)).toHaveLength(3);
+    expect(screen.queryAllByText("delete")).toHaveLength(0);
+
+    await userEvent.setup().click(screen.getByText(t("g.submit")));
+  });
+
+  it("should include both value files and newly added files in submit data for UploadList and UploadDragger", async () => {
```
aktasmehmet: We can delete this test. Actually this is a component-internal functionality which you've already put new tests in the components themselves.

aktasmehmet (changes requested):
Great work 💪  There are really scenarios that we missed completely, in terms of testing and functionality.

But, there are also some things to improve in the PR which I've noted 😉

src/lib/components/Upload/UploadDragger/UploadDragger.test.tsx
atakanbayrak: Did you open task for it if not maybe @emreekiziltoprak could do it and we could look up for it in advance ?

src/lib/components/Form/test/Form.test.tsx
```diff
@@ -324,13 +324,27 @@ describe("Form", () => {
...
+        {/* Upload Components */}
+        <Form.Field name="uploadInput" disabled>
+          <UploadInput {...requiredProps} value={[serverFile]} />
+        </Form.Field>
+        <Form.Field name="uploadList" disabled>
+          <UploadList {...requiredProps} value={[serverFile]} />
+        </Form.Field>
+        <Form.Field name="uploadDragger" disabled>
+          <UploadDragger {...requiredProps} value={[serverFile]} />
+        </Form.Field>
```
aktasmehmet: You should move all new upload related items into "groupItems" const above.

src/lib/components/Form/test/Form.test.tsx
```diff
@@ -419,22 +434,42 @@ describe("Form", () => {
...
+        {/* Upload Components */}
+        <Form.Field name="uploadInput" readOnly>
+          <UploadInput {...requiredProps} value={[serverFile]} />
+        </Form.Field>
+        <Form.Field name="uploadList" readOnly>
+          <UploadList {...requiredProps} value={[serverFile]} />
+        </Form.Field>
+        <Form.Field name="uploadDragger" readOnly>
+          <UploadDragger {...requiredProps} value={[serverFile]} />
+        </Form.Field>
```
aktasmehmet: You should move all new upload related items into "groupItems" const above.

src/lib/components/Form/test/Form.test.tsx
```diff
@@ -419,22 +434,42 @@ describe("Form", () => {
   });
 
   it("should prevent typing or editing all items in FormFieldGroup when readOnly prop is given", async () => {
+    const serverFile = { id: "server-1", name: "server-doc.pdf", type: "application/pdf", size: 2048 };
     const handleSubmit = (data: FormSubmitData) => {
-      expect(JSON.stringify(data)).toBe(JSON.stringify(expectedSubmitResponse));
+      expect(data.isValid).toBe(true);
```
aktasmehmet: This is not a right assertion for this test. Think that. User can put an invalid data into a form item and then submit. Data should be invalid even if they are all readonly, right? So, this line is not correct here and should be removed.

src/lib/components/Form/test/Form.test.tsx
```diff
@@ -419,22 +434,42 @@ describe("Form", () => {
...
+      expect(JSON.stringify(data.values.testGroup)).toBe(JSON.stringify(expectedSubmitResponse.values.testGroup));
+      // Upload Components: readOnly fields are still included in submit (unlike disabled), each with its value file.
+      [data.values.uploadInput, data.values.uploadList, data.values.uploadDragger].forEach(files => {
+        expect(files as FileType[]).toHaveLength(1);
+      });
```
aktasmehmet: You already check this above, right? In terms of value. Showing the item in the upload component is not the concern of the form. Besides, this is already tests in Upload components' tests. 

So, this lines are redundant. ```expect(JSON.stringify(data.values.testGroup)).toBe(JSON.stringify(expectedSubmitResponse.values.testGroup));``` already covers upload values.

src/lib/components/Form/test/Form.test.tsx
```diff
@@ -419,22 +434,42 @@ describe("Form", () => {
...
       </Form>,
     );
 
     const user = userEvent.setup();
     const value = "Entered Text";
     const inputItems = getAllInputItems();
 
+    // Upload Components: value files stay visible but their delete actions are hidden.
+    expect(screen.queryAllByText(serverFile.name)).toHaveLength(3);
+    expect(screen.queryAllByText("delete")).toHaveLength(0);
```
aktasmehmet: Those are upload components' internals. Should not be here and they are already tested in upload components. These lines should be removed

src/lib/components/Form/test/Form.test.tsx
```diff
@@ -323,13 +324,27 @@ describe("Form", () => {
...
```
aktasmehmet: Those are upload components' internals. Should not be here and they are already tested in upload components. These lines should be removed

src/lib/components/Upload/UploadProvider.tsx
```diff
@@ -128,6 +129,7 @@ export const UploadProvider = ({ children, props, isUploadInput, size = "md", na
       const request = new XMLHttpRequest();
       activeRequestsRef.current.add(request);
 
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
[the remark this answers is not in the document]
aktasmehmet: +1, this is not necessary

src/lib/components/Upload/UploadProvider.tsx
```diff
@@ -142,6 +144,7 @@ export const UploadProvider = ({ children, props, isUploadInput, size = "md", na
         stallState.timer = setTimeout(() => {
           stallState.stalled = true;
           request.abort();
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
```
aktasmehmet: this is not necessary. If this is needed for you to develop locally and it is redundant here, then you need to check for your IDE's eslint settings.

src/lib/components/Upload/UploadProvider.tsx
```diff
@@ -293,6 +296,26 @@ export const UploadProvider = ({ children, props, isUploadInput, size = "md", na
...
+  const isFirstValueSync = useRef(true);
+
+  // Keeps `selectedFiles` in sync with the caller-controlled `value` prop after mount
+  useEffect(() => {
+    if (isFirstValueSync.current) {
+      isFirstValueSync.current = false;
+      return;
+    }
+    setSelectedFiles(prev => {
+      const previousValueFilesById = new Map(prev.filter(f => f.addedByValue).map(f => [f.id, f]));
```
aktasmehmet: This filter might be redundant. This effect is already triggered by the ``valueEqualityString``which means ``value`` prop is changed. So you can directly use that because value means it is given my the user (what this filter does.)

src/lib/components/Upload/UploadProvider.tsx
```diff
@@ -293,6 +296,26 @@ export const UploadProvider = ({ children, props, isUploadInput, size = "md", na
...
+      const nextValueFiles = (value ?? []).map(valueFile => {
+        const previous = previousValueFilesById.get(valueFile.id);
+        return previous && (previous.deleting || previous.status === STATUS.DELETE_FAIL) ? previous : valueFile;
+      });
+      const localFiles = prev.filter(f => !f.addedByValue);
+      return maxFile > 1 ? [...nextValueFiles, ...localFiles] : nextValueFiles;
```
aktasmehmet, line 312: I think we should discuss here as a team. I am not sure about the scenarios about keeping user given value and uploaded file sync. Should we keep them both? Should we discard any of them? Should we not allow? etc. I guess it will be perfect if you can arrange a meeting for this 🙏

aktasmehmet (changes requested):

src/lib/components/Form/test/testHelper.tsx
```diff
@@ -42,6 +45,30 @@ export const expectedSubmitResponse = {
       inputDate: DateUtils.getTodayTimeless(),
       switch: false,
       pinCode: ["a", "b"],
+      uploadInput: [
+        {
+          id: serverFile.id,
+          file: { name: serverFile.name, type: serverFile.type, size: serverFile.size },
+          status: STATUS.SUCCESS,
+          uploaded: true,
+        },
```
aktasmehmet: Maybe you can put this in a const as uploadExpectedSubmitResponse or something? You've used the exact same object 3 times here.

src/lib/components/Upload/UploadProvider.tsx
aktasmehmet: I've created as task for this: EDKUI-1491.

Resolving for now.

aktasmehmet (approved):

atakanbayrak (approved):

PR #106, opened 2026-07-20
fix(ImageUpload): implement value prop

ZehranurC:
## Description

- Add value prop to ImageUpload to display images from database. 
- Support src prop for preview URLs instead of File objects
- Image component renders database images with thumbnail, delete, and preview functionality
- Add WithDatabaseImage story and three comprehensive tests

## Type of Change

- [ ] 🆕 New Component
- [x] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Checklist

## How to Test

```
<ImageUpload
  uploadRequest={...}
  deleteRequest={...}
  value={[
    {
      id: "db-1",
      name: "existing.jpg",
      size: 102400,
      type: "image/jpeg",
      src: "https://example.com/image.jpg"
    }
  ]}
/>

```


src/lib/components/Upload/ImageUpload/components/Image.tsx
```diff
@@ -10,15 +9,17 @@ import Preview from "@/components/Upload/ImageUpload/components/Preview";
...
 };
 
-export const Image = ({ file: { status, progress, file } }: Props) => {
+export const Image = ({ file: { status, progress, file, src, addedByValue } }: Props) => {
   const { selectedFiles, removeFiles } = useContext(UploadContext);
-  const image = URL.createObjectURL(file as File);
+  const image = src || (file instanceof File ? URL.createObjectURL(file) : undefined);
   const { visible, show, hide } = useToggle(false);
-  const failed = status === STATUS.CHECK_FAIL || status === STATUS.UPLOAD_FAIL;
+  const failed = status === STATUS.CHECK_FAIL || status === STATUS.UPLOAD_FAIL || (addedByValue && !src);
```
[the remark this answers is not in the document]
atakanbayrak: Forgotten test case is referred in here, thanks to Claude. 😁

src/lib/components/Upload/ImageUpload/ImageUpload.stories.tsx
```diff
@@ -20,3 +25,18 @@ export default meta;
...
+        id: "db-1",
+        name: "profile.jpg",
+        size: 204800,
+        type: "image/jpeg",
+        src: "https://picsum.photos/200/300",
+      },
+    ],
+    deleteRequest: { url: "https://httpbin.org/delete", method: "DELETE" },
+  },
+};
```
atakanbayrak: We could rename story as WithBeforehandValue, because it does not have to be come from Database, source is not known.

src/lib/components/Upload/ImageUpload/ImageUpload.tsx
```diff
@@ -5,11 +5,14 @@ import ImageDragArea from "@/components/Upload/ImageUpload/components/ImageDragA
 import { MIME_TYPES } from "../constants";
 import { ImageUploadProps } from "./types";
 import usePropsWithThemeDefaults from "../../../motif/hooks/usePropsWithThemeDefaults";
+import { mapExternalValue } from "@/components/Upload/helper";
 
 const ImageUpload = (props: PropsWithRef<ImageUploadProps, HTMLDivElement>) => {
-  const { size, style, ref, className, ...rest } = usePropsWithThemeDefaults("ImageUpload", props);
+  const { size, style, ref, className, value: externalValue, ...rest } = usePropsWithThemeDefaults("ImageUpload", props);
+  const mappedValue = mapExternalValue(externalValue);
```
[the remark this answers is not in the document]
atakanbayrak: +99

src/lib/components/Upload/ImageUpload/components/Image.tsx
```diff
@@ -1,7 +1,6 @@
 import styles from "../ImageUpload.module.scss";
```
atakanbayrak, line 1: Why do we have to pass size and type from value prop, even if it is necessary for component work mechanism we should consider something else because no one would like to write down how much or which type of file do they given it.

atakanbayrak (changes requested):

src/lib/components/Upload/ImageUpload/ImageUpload.tsx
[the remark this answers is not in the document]
aktasmehmet: This is the same issue that we are talking about in @emreekiziltoprak upload value test PR. You may move in parallel

src/lib/components/Upload/ImageUpload/types.ts
```diff
@@ -1,6 +1,8 @@
-import { UploadProps } from "../types";
+import { UploadProps, FileObject } from "../types";
 import { InputCommonProps } from "../../Form/types";
 
 export type ImageUploadProps = Omit<UploadProps, "maxFile" | "accept" | "autoUpload"> & ImageUploadDefaultableProps;
 
-export type ImageUploadDefaultableProps = Pick<InputCommonProps, "size">;
+export type ImageUploadDefaultableProps = Pick<InputCommonProps, "size"> & {
+  value?: FileObject[];
```
aktasmehmet: Why is value in defaultable props? It is the most un-defaultable prop I guess :)

src/lib/components/Upload/ImageUpload/components/Image.tsx
```diff
@@ -10,15 +9,17 @@ import Preview from "@/components/Upload/ImageUpload/components/Preview";
...
+
 type Props = {
   file: FileType;
 };
 
-export const Image = ({ file: { status, progress, file } }: Props) => {
+export const Image = ({ file: { status, progress, file, src, addedByValue } }: Props) => {
   const { selectedFiles, removeFiles } = useContext(UploadContext);
-  const image = URL.createObjectURL(file as File);
+  const image = src || (file instanceof File ? URL.createObjectURL(file) : undefined);
```
aktasmehmet: Why not keeping the previous one like:

```suggestion
  const image = src || URL.createObjectURL(file as File);
```

?

src/lib/components/Upload/ImageUpload/components/Image.tsx
aktasmehmet: "Triggering the fail state if src is not provided" is not a good solution. Fail state represents the state that upload is tried and failed. All of the messages, etc. fits to that. You may check the ss here. I did not upload anything, just src is missing but I got an upload error:

Image

It would be better to remove all the fail state triggers all around this PR and do not create the file at all if src is not provided. So this will show just the uploader at all.

src/lib/components/Upload/ImageUpload/components/ImageDragArea.tsx
```diff
@@ -44,7 +44,11 @@ const ImageDragArea = (props: PropsWithRef<unknown, HTMLDivElement>) => {
           ? t(MESSAGE.DELETE_ERROR)
           : t(MESSAGE.UPLOAD_ERROR),
     );
-  const failed = status === STATUS.CHECK_FAIL || status === STATUS.UPLOAD_FAIL || status === STATUS.DELETE_FAIL;
+  const failed =
+    status === STATUS.CHECK_FAIL ||
+    status === STATUS.UPLOAD_FAIL ||
+    status === STATUS.DELETE_FAIL ||
+    (selectedImage && selectedImage.addedByValue && !selectedImage.src);
```
aktasmehmet: I wrote a comment in Image.tsx file about fail states. You may refer to that.

src/lib/components/Upload/ImageUpload/ImageUpload.test.tsx
```diff
@@ -715,4 +715,68 @@ describe("ImageUpload", () => {
     expect(screen.queryByText(/Only.*format files can be uploaded/)).not.toBeInTheDocument();
     expect(getThumbnail()).not.toBeInTheDocument();
   });
+
+  it("should display database image as thumbnail when using value prop", async () => {
```
aktasmehmet:
```suggestion
  it("should display the value-given image", async () => {
```

src/lib/components/Upload/ImageUpload/ImageUpload.test.tsx
```diff
@@ -715,4 +715,68 @@ describe("ImageUpload", () => {
...
+        ]}
+      />,
+    );
+    await waitFor(() => {
+      expect(getThumbnail()).toBeInTheDocument();
+      expect(getThumbnail()).toHaveAttribute("src", dbSrc);
+    });
+  });
+
+  it("should open preview for database image", async () => {
```
aktasmehmet:
```suggestion
  it("should show the preview of the value-given image", async () => {
```

src/lib/components/Upload/ImageUpload/ImageUpload.test.tsx
```diff
@@ -715,4 +715,68 @@ describe("ImageUpload", () => {
...
+          },
+        ]}
+      />,
+    );
+    await userEvent.click(screen.getByText("visibility"));
+    const previewImg = screen.queryByAltText("Image Preview")!;
+    await waitFor(() => expect(previewImg).toBeInTheDocument());
+  });
+
+  it("should delete database image when delete button is clicked", async () => {
```
aktasmehmet:
```suggestion
  it("should delete the value-given image when delete button is clicked", async () => {
```

src/lib/components/Upload/ImageUpload/ImageUpload.test.tsx
```diff
@@ -715,4 +715,68 @@ describe("ImageUpload", () => {
...
+            size: 102400,
+            type: "image/jpeg",
+            src: dbSrc,
+          },
+        ]}
+      />,
+    );
+    await userEvent.click(screen.getByText("visibility"));
+    const previewImg = screen.queryByAltText("Image Preview")!;
+    await waitFor(() => expect(previewImg).toBeInTheDocument());
```
aktasmehmet: It would be better to check src for the preview here

src/lib/components/Upload/ImageUpload/ImageUpload.stories.tsx
aktasmehmet: I am not sure about the name but definitely not "database". I've tried "value-given" in tests.

aktasmehmet (changes requested):

src/lib/components/Upload/ImageUpload/components/Image.tsx
ZehranurC: The instanceof check is required for type safety. file can be either File (user uploads) or metadata object (database images). Calling URL.createObjectURL directly on metadata objects causes runtime errors. Current implementation correctly handles both cases.

src/lib/components/Upload/ImageUpload/components/Image.tsx
ZehranurC: Size and type are required because component's core logic depends on them:
- File validation (maxSize check)
- MIME type validation 
- Duplicate file detection
- UI display (file size shown)

src/lib/components/Upload/ImageUpload/components/Image.tsx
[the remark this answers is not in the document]
ZehranurC: This check has been removed

src/lib/components/Upload/ImageUpload/ImageUpload.tsx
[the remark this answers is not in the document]
ZehranurC: -  Added useEffect in UploadProvider to sync selectedFiles when value prop changes
-  Updated all components (ImageUpload, UploadInput, UploadDragger, UploadList) to use normalizeValue for type safety
- Added test for async value changes
-  DB image async loading now works.

atakanbayrak (dismissed):

src/lib/components/Upload/helper.ts
```diff
@@ -4,18 +4,42 @@ import { InputValue } from "@/components/Form/types.ts";
...
-          type: f.type,
-          size: f.size,
-        },
-        status: STATUS.SUCCESS,
-        uploaded: true,
-        download: f.onDownloadClick,
-        addedByValue: true,
-      }))
+    ? (value as FileObject[])
+        .filter(f => f.src)
```
aktasmehmet: This filter is so wrong! This type is used by many upload components and by filtering by src, you will eliminate standard files (not image) to be mapeed. This filter should be removed

src/lib/components/Upload/ImageUpload/types.ts
```diff
@@ -1,6 +1,9 @@
-import { UploadProps } from "../types";
+import { UploadProps, FileObject } from "../types";
 import { InputCommonProps } from "../../Form/types";
 
-export type ImageUploadProps = Omit<UploadProps, "maxFile" | "accept" | "autoUpload"> & ImageUploadDefaultableProps;
+export type ImageUploadProps = Omit<UploadProps, "maxFile" | "accept" | "autoUpload"> &
+  ImageUploadDefaultableProps & {
+    value?: FileObject[];
```
aktasmehmet: I am not sure ImageUpload takes multiple values?

src/lib/components/Upload/helper.ts
```diff
@@ -4,18 +4,42 @@ import { InputValue } from "@/components/Form/types.ts";
...
+          },
+          status: STATUS.SUCCESS,
+          uploaded: true,
+          download: f.onDownloadClick,
+          addedByValue: true,
+          src: f.src,
+        }))
+    : [];
+
+export const normalizeValue = (value?: InputValue) =>
```
aktasmehmet: I didnt understand the need for this function.
1 - ``value as (FileType | FileObject)``is used but this is the files coming from outside. They are using FileObject. Why is FileType there?
2 - There is no need for ``if ("file" in f) return f;`` because this should process FileObject

Considering those 2 above, this method becomes the same with ``mapExternalValue``. So it is not needed. What am I missing?

src/lib/components/Upload/UploadInput/UploadInput.tsx
```diff
@@ -14,7 +15,7 @@ import { UploadInputProps, UploadInputWrapperProps } from "./types";
 import usePropsWithThemeDefaults from "../../../motif/hooks/usePropsWithThemeDefaults";
 import { sanitizeModuleRootClasses } from "../../../../utils/cssUtils";
 import { MotifIcon } from "@/components/Motif/Icon";
-import { mapExternalValue, toFormValue } from "@/components/Upload/helper.ts";
+import { toFormValue } from "@/components/Upload/helper.ts";
```
aktasmehmet: Why dont you put seperate but same "from" here? You can import them together?

src/lib/components/Upload/ImageUpload/components/Image.tsx
aktasmehmet: Dont you map all to the same type with mapping or normalizing?

src/lib/components/Upload/UploadProvider.tsx
```diff
@@ -3,14 +3,15 @@ import { createContext, useCallback, useEffect, useRef, useState } from "react";
...
+import { normalizeValue } from "@/components/Upload/helper";
 
 export const UploadContext = createContext<UploadContextType>(ContextDefaultValues);
 
 export const UploadProvider = ({ children, props, isUploadInput, size = "md", name, disabled, value }: UploadProviderProps) => {
   const { maxFile = 1, autoUpload = true, messages, uploadRequest, deleteRequest, maxSize, accept, customValidation } = props;
   const hiddenInputRef = useRef<HTMLInputElement>(null);
   const activeRequestsRef = useRef<Set<XMLHttpRequest>>(new Set());
-  const [selectedFiles, setSelectedFiles] = useState<FileType[]>(value ?? []);
+  const [selectedFiles, setSelectedFiles] = useState<FileType[]>(normalizeValue(value));
```
aktasmehmet: You already pass normalized value to this Provider? I dont think this change is necessary.

src/lib/components/Upload/ImageUpload/components/Image.tsx
aktasmehmet: There is still an issue here. I didnt give the src but this time it behaves like there is an image.

**No src -> Initial state as no file at all**

Screenshot 2026-07-29 at 10 43 00

Simple is that. Please pay extra attention to test here

aktasmehmet: And please please please add a test for that

src/lib/components/Upload/UploadProvider.tsx
```diff
@@ -362,6 +363,15 @@ export const UploadProvider = ({ children, props, isUploadInput, size = "md", na
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [accept, autoUpload, isUploadInput, maxFile, maxSize, messages, selectedFilesEqualityString, uploadV2]);
 
+  // Sync value-based files when value prop changes
+  useEffect(() => {
```
aktasmehmet: There are too many issue Claude mentioned here. More importantly there is a better approach (and you will definitely get hit by a conflict) by @emreekiziltoprak 's solution here: https://github.com/motif-ui/motifreact/pull/105/changes#diff-1a320c016d03d091f61b5fedfed44739d513482cf9621fcbf0baf870b469ca73R297

I guess you should wait and update here. Maybe there is no need at all.

aktasmehmet (changes requested):

src/lib/components/Upload/ImageUpload/components/Image.tsx
aktasmehmet: This still continue in a different way. Now, when src is wrong and not loading an actual image, we have the same issue. 

I know we discarded this before when there are no value prop but now the scenario is different.

The clearest way I think yo put this into React rendering process and then catch the issue on ``onError`` of **** element. Like change the image const in Image.tsx as a state object and when component is first loaded, set state value there. This time, we may see a little flicker but you can catch the issue on onError of img component. Then you can remove the buttons and even you can show a backup image there, instead of a broken image.

The other option is to use ref for  element and register an onError event for that ref and then catch onError.

The reason for all these is that if you give src directly, it will not load the image, then the onError callback will execute. Basically it wont :)

src/lib/components/Upload/ImageUpload/components/Image.tsx
```diff
@@ -10,29 +9,28 @@ import Preview from "@/components/Upload/ImageUpload/components/Preview";
...
+
 type Props = {
   file: FileType;
 };
 
-export const Image = ({ file: { status, progress, file, deleting } }: Props) => {
+export const Image = ({ file: { status, progress, file, src, deleting } }: Props) => {
   const { selectedFiles, removeFiles } = useContext(UploadContext);
-  const image = URL.createObjectURL(file as File);
+  const image = src || (file instanceof File ? URL.createObjectURL(file) : undefined);
```
aktasmehmet: I put here.

There is a bug now. When the button is clicked, it opens the file browser:

https://github.com/user-attachments/assets/71d86f12-a343-42d4-9504-b21f90968af1

src/lib/components/Upload/ImageUpload/ImageUpload.stories.tsx
```diff
@@ -20,3 +25,16 @@ export default meta;
 type Story = StoryObj<typeof ImageUpload>;
 
 export const Primary: Story = {};
+
+export const WithValueGivenImage: Story = {
```
aktasmehmet: I dont think we need a seperate story for that. nothing special. Like other props. Give the prop and render something.

So you can remove this.

aktasmehmet (changes requested):

src/lib/components/Upload/ImageUpload/components/Image.tsx
```diff
@@ -1,37 +1,40 @@
...
-      variant="danger"
-      className={styles.icon}
-      disabled={deleting}
-      onClick={() => removeFiles([selectedFiles[0]])}
-    />
-  );
+  useEffect(() => {
+    const image = addedByValue ? src : file instanceof File ? URL.createObjectURL(file) : undefined;
+    image ? setImage(image) : setMaybeBrokenSrc(true);
+  }, [addedByValue, file, src]);
```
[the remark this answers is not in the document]
aktasmehmet: this is not necessary

aktasmehmet (approved):

src/lib/components/Upload/ImageUpload/components/Image.tsx
```diff
@@ -1,37 +1,46 @@
...
+  useEffect(() => {
+    const objectUrl = !addedByValue && file instanceof File ? URL.createObjectURL(file) : undefined;
+    const image = addedByValue ? src : objectUrl;
+
+    image ? setImage(image) : setMaybeBrokenSrc(true);
 
+    return () => {
+      objectUrl && URL.revokeObjectURL(objectUrl);
+    };
+  }, [addedByValue, file, src]);
```
ZehranurC, line 36:
```suggestion
  }, []);
```

aktasmehmet: no, it is put there on purpose. If not, file changes cannot be detected

ZehranurC: Sorry, I misunderstood. The effect must detect image changes.

atakanbayrak (approved):

PR #109, opened 2026-07-21
fix(InputText): text transform feature is added

atakanbayrak:
## Description

Adds a textTransform prop to InputText (undefined | "uppercase" | "lowercase" | "capitalize") that converts the input's value on every keystroke.

## Type of Change

- [ ] 🆕 New Component
- [x] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Screenshots / Preview

Screenshot 2026-07-21 at 10 56 32

## Checklist

## How to Test

```code
      <InputText textTransform="uppercase" />
```

### DIFF
--- a/src/lib/components/InputText/InputText.test.tsx
+++ b/src/lib/components/InputText/InputText.test.tsx
@@ -115,4 +115,14 @@ describe("InputText", () => {
     const { getByTestId } = render(<InputText pill />);
     expect(getByTestId("inputItem")).toHaveClass("pill");
   });
+
+  it("should apply textTransform without moving the caret when editing mid-string", async () => {
+    render(<InputText textTransform="uppercase" value="ABC" onChange={jest.fn()} placeholder="Test" />);
+    const input = screen.getByDisplayValue<HTMLInputElement>("ABC");
+    await userEvent.type(input, "x", { initialSelectionStart: 1, initialSelectionEnd: 1 });
+
+    expect(input).toHaveValue("AXBC");
+    expect(input.selectionStart).toBe(2);
+    expect(input.selectionEnd).toBe(2);
+  });
 });
--- a/src/lib/components/InputText/InputText.test.tsx
+++ b/src/lib/components/InputText/InputText.test.tsx
@@ -3,6 +3,8 @@ import InputText from "@/components/InputText/InputText";
 import { userEvent } from "@testing-library/user-event";
 import { InputSize } from "../Form/types";
 import Icon from "@/components/Icon";
+import MotifProvider from "../../motif/context/MotifProvider";
+import { TextTransform } from "@/components/Motif/InputText/types.ts";
 
 describe("InputText", () => {
   it("should be rendered with only required props", () => {
@@ -124,5 +126,44 @@ describe("InputText", () => {
     expect(input).toHaveValue("AXBC");
     expect(input.selectionStart).toBe(2);
     expect(input.selectionEnd).toBe(2);
+  });
+
+  it("should transform typed value according to the textTransform prop", async () => {
+    const cases: { textTransform: TextTransform; typed: string; expected: string }[] = [
+      { textTransform: "uppercase", typed: "hello world", expected: "HELLO WORLD" },
+      { textTransform: "lowercase", typed: "HELLO WORLD", expected: "hello world" },
+      { textTransform: "capitalize", typed: "hello world", expected: "Hello World" },
+    ];
+
+    for (const { textTransform, typed, expected } of cases) {
+      const handleChange = jest.fn();
+      const { unmount } = render(<InputText textTransform={textTransform} value="" onChange={handleChange} placeholder="Test" />);
+      const input = screen.getByPlaceholderText("Test");
+      await userEvent.type(input, typed);
+
+      expect(input).toHaveValue(expected);
+      unmount();
+    }
+  });
+
+  it("should apply locale-aware textTransform when locale is set to tr", async () => {
+    const cases: { textTransform: TextTransform; typed: string; expected: string }[] = [
+      { textTransform: "uppercase", typed: "istanbul", expected: "İSTANBUL" },
+      { textTransform: "lowercase", typed: "ISTANBUL", expected: "ıstanbul" },
+      { textTransform: "capitalize", typed: "istanbul", expected: "İstanbul" },
+    ];
+
+    for (const { textTransform, typed, expected } of cases) {
+      const { unmount } = render(
+        <MotifProvider locale="tr">
+          <InputText textTransform={textTransform} value="" onChange={jest.fn()} placeholder="Test" />
+        </MotifProvider>,
+      );
+      const input = screen.getByPlaceholderText("Test");
+      await userEvent.type(input, typed);
+
+      expect(input).toHaveValue(expected);
+      unmount();
+    }
   });
 });
--- a/src/lib/components/InputText/InputText.test.tsx
+++ b/src/lib/components/InputText/InputText.test.tsx
@@ -128,21 +128,42 @@ describe("InputText", () => {
     expect(input.selectionEnd).toBe(2);
   });
 
-  it("should transform typed value according to the textTransform prop", async () => {
-    const cases: { textTransform: TextTransform; typed: string; expected: string }[] = [
-      { textTransform: "uppercase", typed: "hello world", expected: "HELLO WORLD" },
-      { textTransform: "lowercase", typed: "HELLO WORLD", expected: "hello world" },
-      { textTransform: "capitalize", typed: "hello world", expected: "Hello World" },
+  it("should apply textTransform consistently for value, typing and value-change scenarios", async () => {
+    const initialValue = "hElLo wORld";
+    const changedValue = "hElLo wORld 2";
+    const cases: { textTransform: TextTransform; expected: string; expectedAfterChange: string }[] = [
+      { textTransform: "uppercase", expected: "HELLO WORLD", expectedAfterChange: "HELLO WORLD 2" },
+      { textTransform: "lowercase", expected: "hello world", expectedAfterChange: "hello world 2" },
+      { textTransform: "capitalize", expected: "HElLo WORld", expectedAfterChange: "HElLo WORld 2" },
     ];
 
-    for (const { textTransform, typed, expected } of cases) {
-      const handleChange = jest.fn();
-      const { unmount } = render(<InputText textTransform={textTransform} value="" onChange={handleChange} placeholder="Test" />);
-      const input = screen.getByPlaceholderText("Test");
-      await userEvent.type(input, typed);
+    for (const { textTransform, expected, expectedAfterChange } of cases) {
+      // 1 - value control
+      const { unmount: unmountValueControl } = render(
+        <InputText textTransform={textTransform} value={initialValue} onChange={jest.fn()} placeholder="Test" />,
+      );
+      expect(screen.getByPlaceholderText("Test")).toHaveValue(expected);
+      unmountValueControl();
 
-      expect(input).toHaveValue(expected);
-      unmount();
+      // 2 - typing control
+      const { unmount: unmountTypingControl } = render(
+        <InputText textTransform={textTransform} value="" onChange={jest.fn()} placeholder="Test" />,
+      );
+      const typingInput = screen.getByPlaceholderText("Test");
+      await userEvent.type(typingInput, initialValue);
+      expect(typingInput).toHaveValue(expected);
+      unmountTypingControl();
+
+      // 3 - value change control (rerender)
+      const { rerender, unmount: unmountValueChangeControl } = render(
+        <InputText textTransform={textTransform} value={initialValue} onChange={jest.fn()} placeholder="Test" />,
+      );
+      const rerenderInput = screen.getByPlaceholderText("Test");
+      expect(rerenderInput).toHaveValue(expected);
+
+      rerender(<InputText textTransform={textTransform} value={changedValue} onChange={jest.fn()} placeholder="Test" />);
+      expect(rerenderInput).toHaveValue(expectedAfterChange);
+      unmountValueChangeControl();
     }
   });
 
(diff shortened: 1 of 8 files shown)


src/lib/components/InputText/types.ts
```diff
@@ -6,6 +6,7 @@ export type InputTextDefaultableProps = {
   pill?: boolean;
```
aktasmehmet, line 6: Image

Can you fix the title to prevent duplicate "InputText"? Thx.

aktasmehmet (changes requested):

src/lib/components/Motif/InputText/helper.ts
```diff
@@ -0,0 +1,12 @@
+export const applyTextTransform = (value: string, textTransform: string) => {
+  switch (textTransform) {
+    case "uppercase":
+      return value.toUpperCase();
+    case "lowercase":
+      return value.toLowerCase();
```
aktasmehmet: ``locale`` variants of those functions should be used.

src/lib/components/Motif/InputText/helper.ts
```diff
@@ -0,0 +1,12 @@
+export const applyTextTransform = (value: string, textTransform: string) => {
```
aktasmehmet: textTransform type should be defined as a combined type instead of string.

src/lib/components/Motif/InputText/InputText.tsx
```diff
@@ -77,13 +80,13 @@ const InputText = (props: PropsWithRef<InternalInputProps, HTMLDivElement>) => {
...
       }
-      !uncontrolled && setItemValue(val);
-      onChange?.(val);
-      if (updateInputRefValue && inputRef.current) {
-        inputRef.current.value = val;
+      !uncontrolled && setItemValue(transformedVal);
+      onChange?.(transformedVal);
+      if ((updateInputRefValue || textTransform) && inputRef.current) {
+        inputRef.current.value = transformedVal;
       }
```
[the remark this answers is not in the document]
aktasmehmet: +1

src/lib/components/InputText/types.ts
```diff
@@ -6,6 +6,7 @@ export type InputTextDefaultableProps = {
...
   size?: InputSize;
   clearable?: boolean;
+  textTransform?: "uppercase" | "lowercase" | "capitalize";
```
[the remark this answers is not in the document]
aktasmehmet: +1, especially for the test. I was going to write. You should write the related test. Storybook action is not necessary because it'll handle the new prop automatically.

src/lib/components/Motif/InputText/InputText.tsx
```diff
@@ -63,8 +65,9 @@ const InputText = (props: PropsWithRef<InternalInputProps, HTMLDivElement>) => {
 
   const changeProcess = useCallback(
     (val: string, updateInputRefValue?: boolean) => {
+      const transformedVal = textTransform ? applyTextTransform(val, textTransform) : val;
```
aktasmehmet: I would go with this to not make too many name related changes. Besides, the value is not always transformed so name convention is not that good I think.


```suggestion
(typedVal: string, updateInputRefValue?: boolean) => {
      const val = textTransform ? applyTextTransform(typedVal, textTransform) : typedVal;
```

src/lib/components/Motif/InputText/InputText.tsx
```diff
@@ -77,13 +80,13 @@ const InputText = (props: PropsWithRef<InternalInputProps, HTMLDivElement>) => {
         if (inputRef.current) inputRef.current.value = processed;
         return;
       }
-      !uncontrolled && setItemValue(val);
-      onChange?.(val);
-      if (updateInputRefValue && inputRef.current) {
-        inputRef.current.value = val;
+      !uncontrolled && setItemValue(transformedVal);
+      onChange?.(transformedVal);
+      if ((updateInputRefValue || textTransform) && inputRef.current) {
```
aktasmehmet: I think this is wrong. This may pass the condition even if updateInputRefValue is false. Which is the main reason that prop is there. In fact, I did not understand the purpose here. It looks redundant and to be reverted.

aktasmehmet (changes requested):

src/lib/components/Motif/InputText/helper.ts
```diff
@@ -0,0 +1,15 @@
...
+import { capitalizeFirstLetter } from "src/utils/utils.ts";
+
+export const applyTextTransform = (value: string, textTransform: TextTransform) => {
+  switch (textTransform) {
+    case "uppercase":
+      return value.toUpperCase();
+    case "lowercase":
+      return value.toLowerCase();
+    case "capitalize":
+      return capitalizeFirstLetter(value);
```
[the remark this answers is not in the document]
aktasmehmet: I did not check completely. @atakanbayrak  is this true? Capitalize should make the first letter of EVERY word, not just the first word. Can you check this?

src/lib/components/InputText/types.ts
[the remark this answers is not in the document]
aktasmehmet: The test you've added is caret related. There has to be test to check all 3 values of the prop. Our very basic test :(

src/lib/components/Motif/InputText/helper.ts
aktasmehmet: You haven't given the locale coming from the context, as parameter?

aktasmehmet (changes requested):
There are still too much to change. I'll wait for them to be finished before checking the ref and caret thing.

src/lib/components/InputText/InputText.test.tsx
```diff
@@ -115,4 +117,53 @@ describe("InputText", () => {
...
+      const { unmount } = render(<InputText textTransform={textTransform} value="" onChange={handleChange} placeholder="Test" />);
+      const input = screen.getByPlaceholderText("Test");
+      await userEvent.type(input, typed);
+
+      expect(input).toHaveValue(expected);
+      unmount();
+    }
+  });
+
+  it("should apply locale-aware textTransform when locale is set to tr", async () => {
```
aktasmehmet:
```suggestion
  it("should apply textTransform locale-aware based on the locale value of the lib", async () => {
```
You've tested TR but this is not actually TR related, so test name should be changed

src/lib/components/Motif/InputText/helper.ts
```diff
@@ -0,0 +1,22 @@
...
+      return value.replace(/(^|\s)\S/g, char => char.toLocaleUpperCase(locale));
+    default:
+      return value;
+  }
+};
+
+export const useTextTransform = () => {
+  const { locale } = useMotifContext();
+  return useCallback((value: string, textTransform: TextTransform) => applyTextTransform(value, textTransform, locale), [locale]);
+};
```
aktasmehmet: Is there a special reason that you've created a hook, instead of just calling applyTextTransform in the necessary component by calling locale from context, there?

There are some issues:

1 - You use a hook in a helper file,
2 - Unnecessary hook usage :)

src/lib/components/Motif/InputText/InputText.tsx
```diff
@@ -40,19 +41,34 @@ const InputText = (props: PropsWithRef<InternalInputProps, HTMLDivElement>) => {
     onClearClick,
     onValueUpdated,
     valueTransformer,
+    textTransform,
```
aktasmehmet: it is not working when value is given instead of typing:

NOTE: Do not forget to add tests for value options like this, for all possbile transform types

```<InputText size="lg" textTransform="capitalize" value="mehmet aktaş" />```

Image

File: src/lib/components/Motif/InputText/InputText.tsx [hunk 12]
```diff
@@ -40,19 +41,34 @@ const InputText = (props: PropsWithRef<InternalInputProps, HTMLDivElement>) => {
...
+  // characters (e.g. lowercase -> uppercase), even though the length is unchanged. Restore
+  // the caret right after that DOM write commits, before the browser paints.
+  useLayoutEffect(() => {
+    if (pendingSelectionRef.current && inputRef.current) {
+      const { start, end } = pendingSelectionRef.current;
+      inputRef.current.setSelectionRange(start, end);
+      pendingSelectionRef.current = null;
+    }
+  }, [itemValue]);
+
```
aktasmehmet: Now, forget about all these, there is much more easier way. When value changes, just keep track of the cursor and after the update, move that back.

Revert everything in changeProcess here and try this:

```
const changeProcess = useCallback(
    (typedValue: string, updateInputRefValue?: boolean) => {
      const inputEl = inputRef.current;

      const transformerToUse = textTransform ? (v: string) => applyTextTransform(v, textTransform) : valueTransformer;

      if (transformerToUse) {
        // 1. Save current cursor positions
        const start = inputEl?.selectionStart ?? null;
        const end = inputEl?.selectionEnd ?? null;

        const processed = transformerToUse(typedValue);
        if (processed === undefined) {
          if (inputEl) inputEl.value = itemValue;
          return;
        }
        !uncontrolled && setItemValue(processed);
        onChange?.(processed);
        if (inputEl) {
          inputEl.value = processed;
          // 2. Restore cursor position if input value was modified
          if (start !== null && end !== null) {
            inputEl.setSelectionRange(start, end);
          }
        }
        return;
      }

      !uncontrolled && setItemValue(typedValue);
      onChange?.(typedValue);

      if (updateInputRefValue && inputEl) {
        inputEl.value = typedValue;
      }
    },
    [onChange, setItemValue, uncontrolled, valueTransformer, itemValue, textTransform],
  );
```

I tried and works like a charm.

In fact, I got another idea comes to my mind. Since this component now both supports numbers and string, we should not put a direct string function here. In fact, thats the reason why I put the transformer prop. What do you think about moving this textTransform out of here (by just keeping the cursor thing) and give that as a transformer from InputText and InputPassword, etc?

aktasmehmet (changes requested):

src/lib/components/Motif/InputText/helper.ts
atakanbayrak: Yes because I don't want to pass through locale prop via InputText to applyTextTransform. If we need to have this method somewhere else in the future we don't have to pass locale prop through component to helper function.

src/lib/components/Motif/InputText/InputText.tsx
```diff
@@ -50,37 +50,48 @@ const InputText = (props: PropsWithRef<InternalInputProps, HTMLDivElement>) => {
   const prevValueRef = useRef(value);
   useImperativeHandle(imperativeRef, () => ({ valueStateSetter: setItemValue }));
 
-  const [itemValue, setItemValue] = useState(value);
+  const [itemValue, setItemValue] = useState(() => (valueTransformer ? (valueTransformer(value) ?? value) : value));
```
aktasmehmet:
```suggestion
  const [itemValue, setItemValue] = useState(() => valueTransformer?.(value) ?? value);
```

src/lib/components/Motif/InputText/InputText.tsx
```diff
@@ -50,37 +50,48 @@ const InputText = (props: PropsWithRef<InternalInputProps, HTMLDivElement>) => {
...
   const controlledProps = uncontrolled ? { defaultValue: value } : { value: itemValue };
-
   useEffect(() => {
     if (!uncontrolled && value !== prevValueRef.current) {
       prevValueRef.current = value;
```
aktasmehmet: you should put this after nextValue and set nextValue

src/lib/components/Motif/InputText/InputText.tsx
```diff
@@ -50,37 +50,48 @@ const InputText = (props: PropsWithRef<InternalInputProps, HTMLDivElement>) => {
...
-      setItemValue(value);
-      onValueUpdated?.(value);
+      const nextValue = valueTransformer ? (valueTransformer(value) ?? value) : value;
```
aktasmehmet:
```suggestion
      const nextValue = valueTransformer?.(value) ?? value;
      prevValueRef.current = nextValue;
```

src/lib/components/InputText/InputText.tsx
```diff
@@ -27,6 +30,7 @@ const InputText = (p: PropsWithRef<InputTextProps, HTMLDivElement>) => {
       {...propsFromForm}
       value={props.value as string}
       onValueUpdated={onFormFieldValueUpdate}
+      valueTransformer={textTransform ? v => applyTextTransform(v, textTransform, locale) : undefined}
```
aktasmehmet:
```suggestion
      valueTransformer={textTransform && (v => applyTextTransform(v, textTransform, locale))}
```

src/lib/components/InputText/InputText.test.tsx
```diff
@@ -115,4 +117,81 @@ describe("InputText", () => {
...
+  });
+
+  it("should apply textTransform to the given value prop", () => {
+    const cases: { textTransform: TextTransform; value: string; expected: string }[] = [
+      { textTransform: "uppercase", value: "hello world", expected: "HELLO WORLD" },
+      { textTransform: "lowercase", value: "HELLO WORLD", expected: "hello world" },
+      { textTransform: "capitalize", value: "hello world", expected: "Hello World" },
+    ];
+
+    for (const { textTransform, value, expected } of cases) {
```
aktasmehmet: This is shat Claude gives :) Cmon, convert it into foreach :)

src/lib/components/InputText/InputText.test.tsx
```diff
@@ -115,4 +117,81 @@ describe("InputText", () => {
...
+      );
+      const input = screen.getByPlaceholderText("Test");
+      await userEvent.type(input, typed);
+
+      expect(input).toHaveValue(expected);
+      unmount();
+    }
+  });
+
+  it("should apply textTransform to the given value prop", () => {
```
aktasmehmet: You may use a single forEach in this test to test them all. Besides, you should also put a value typing check for that. You may use a smarter start point as value to include all scenarios like lowercase, uppercase, etc: ```hElLo wORld```

Test should be:

1 - value control -> hElLo wORld -> correct ?
2 - typing control -> hElLo wORld -> correct?
3 - value change control (rerender) -> hElLo wORld 2 -> correct?

src/lib/components/InputText/InputText.test.tsx
atakanbayrak: Screenshot 2026-07-28 at 11 00 35

I looked for the other examples that we wrote down previously but this is the exact pattern which we use for test cases and this is foreach, only difference is that there are 3 different value that we checked.

aktasmehmet: yes those were before we are using unmount pattern

src/lib/components/InputText/InputText.test.tsx
```diff
@@ -115,4 +117,74 @@ describe("InputText", () => {
     const { getByTestId } = render(<InputText pill />);
     expect(getByTestId("inputItem")).toHaveClass("pill");
   });
+
```
aktasmehmet, line 120: In every test you've put in this document has unnecessary onChange prop in renderen InputText. They should be removed. They are not related with the tests.

src/lib/components/InputText/InputText.test.tsx
```diff
@@ -115,4 +117,74 @@ describe("InputText", () => {
...
+      // 3 - value change control (rerender)
+      const { rerender, unmount: unmountValueChangeControl } = render(
+        <InputText textTransform={textTransform} value={initialValue} onChange={jest.fn()} placeholder="Test" />,
+      );
+      const rerenderInput = screen.getByPlaceholderText("Test");
+      expect(rerenderInput).toHaveValue(expected);
+
+      rerender(<InputText textTransform={textTransform} value={changedValue} onChange={jest.fn()} placeholder="Test" />);
+      expect(rerenderInput).toHaveValue(expectedAfterChange);
+      unmountValueChangeControl();
```
aktasmehmet: You may use clean instead of using unmounts with different names. This might be clearer.


```suggestion
      // 1 - initial value
      render(<InputText textTransform={textTransform} value={initialValue} placeholder="Test" />);
      expect(screen.getByPlaceholderText("Test")).toHaveValue(expected);
      cleanup();

      // 2 - typing
      render(<InputText textTransform={textTransform} value="" placeholder="Test" />);
      const typingInput = screen.getByPlaceholderText("Test");
      await userEvent.type(typingInput, initialValue);
      expect(typingInput).toHaveValue(expected);
      cleanup();

      // 3 - value change (rerender)
      const { rerender } = render(<InputText textTransform={textTransform} value={initialValue} placeholder="Test" />);
      const rerenderInput = screen.getByPlaceholderText("Test");
      expect(rerenderInput).toHaveValue(expected);

      rerender(<InputText textTransform={textTransform} value={changedValue} placeholder="Test" />);
      expect(rerenderInput).toHaveValue(expectedAfterChange);
      cleanup();
```

aktasmehmet (approved):

PR #111, opened 2026-07-21
fix(InputText): InputText and Select's different height problem is solved

atakanbayrak:
## Description

Select and InputText different height problem is solved, main problem was pop up from inputWrapper div wrapper className. 

Secondly, this wrapper class implemented when clear icon tried to implement. Because of this reason, clear icon is placed correctly to component. 

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [x] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Screenshots / Preview

Screenshot 2026-07-21 at 14 45 55
Screenshot 2026-07-21 at 14 41 18

## Checklist

## How to Test

```code
<Form onSubmit={data => alert(JSON.stringify(data))} size="xs">
        <Grid>
          <Grid.Row>
            <Grid.Col>
              <Form.Field name="name" label="Name">
                <InputText value="Predefined name" size="xs" />
              </Form.Field>
            </Grid.Col>
          </Grid.Row>

          <Grid.Row>
            <Grid.Col>
              <Form.Field name="name" label="Name">
                <Select
                  size="sm"
                  data={[
                    { label: "10", value: "10" },
                    { label: "20", value: "20" },
                    { label: "30", value: "30" },
                  ]}
                />
              </Form.Field>
            </Grid.Col>
          </Grid.Row>
        </Grid>
      </Form>

      <InputText size="lg" clearable />
      <Select
        size="lg"
        data={[
          { label: "10", value: "10" },
          { label: "20", value: "20" },
          { label: "30", value: "30" },
        ]}
      />
```

### DIFF
--- a/src/lib/components/Motif/InputText/InputText.module.scss
+++ b/src/lib/components/Motif/InputText/InputText.module.scss
@@ -107,6 +107,7 @@ $sizes: (("xs", "8x", "4x", "18x"), ("sm", "12x", "6x", "20x"), ("md", "16x", "8
 
 .inputWrapper {
   position: relative;
+  flex: 1;
   display: flex;
   min-width: 0;
 }
--- a/src/lib/components/Motif/InputText/InputText.module.scss
+++ b/src/lib/components/Motif/InputText/InputText.module.scss
@@ -18,7 +18,7 @@ $sizes: (("xs", "8x", "4x", "18x"), ("sm", "12x", "6x", "20x"), ("md", "16x", "8
   }
 
   input {
-    flex: 1;
+    display: block;
     box-sizing: border-box;
     border: none;
     outline: none;
@@ -113,7 +113,9 @@ $sizes: (("xs", "8x", "4x", "18x"), ("sm", "12x", "6x", "20x"), ("md", "16x", "8
 
 .clearButton {
   position: absolute;
+  top: 50%;
   right: 0;
+  transform: translateY(-50%);
 }
 
 .loader {
--- a/src/lib/components/Motif/InputText/InputText.module.scss
+++ b/src/lib/components/Motif/InputText/InputText.module.scss
@@ -18,7 +18,6 @@ $sizes: (("xs", "8x", "4x", "18x"), ("sm", "12x", "6x", "20x"), ("md", "16x", "8
   }
 
   input {
-    display: block;
     box-sizing: border-box;
     border: none;
     outline: none;
--- a/src/lib/components/Motif/InputText/InputText.module.scss
+++ b/src/lib/components/Motif/InputText/InputText.module.scss
@@ -107,15 +107,13 @@ $sizes: (("xs", "8x", "4x", "18x"), ("sm", "12x", "6x", "20x"), ("md", "16x", "8
 
 .inputWrapper {
   position: relative;
-  flex: 1;
+  display: flex;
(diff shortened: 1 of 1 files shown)


src/lib/components/Motif/InputText/InputText.module.scss
```diff
@@ -18,7 +18,7 @@ $sizes: (("xs", "8x", "4x", "18x"), ("sm", "12x", "6x", "20x"), ("md", "16x", "8
   }
 
   input {
-    flex: 1;
+    display: block;
```
aktasmehmet: Default diplay property of input is inline-block. I am not a fan of changing this for this easy issue. I've checked input text and select and it is clear that the issue is in ``mtf-InputText--inputWrapper`` Giving this flex or inline fex will resolve this.

Regardless of this, flex:1 and width: %100 both serves the same thing. You may check and get rid of any of them 😉

src/lib/components/Motif/InputText/InputText.module.scss
```diff
@@ -113,7 +113,9 @@ $sizes: (("xs", "8x", "4x", "18x"), ("sm", "12x", "6x", "20x"), ("md", "16x", "8
 
 .clearButton {
   position: absolute;
+  top: 50%;
   right: 0;
+  transform: translateY(-50%);
```
aktasmehmet: Using flex in the wrapper will also fix the fix you've done here :) You may revert these

aktasmehmet (changes requested):

src/lib/components/Motif/InputText/InputText.module.scss
atakanbayrak: When I use flex in here text is overlapped with the clear icon. Did you try it on that way or was it just an idea ?

atakanbayrak: I implemented without using flex:1 and everything seems OK.

src/lib/components/Motif/InputText/InputText.module.scss
```diff
@@ -108,6 +108,7 @@ $sizes: (("xs", "8x", "4x", "18x"), ("sm", "12x", "6x", "20x"), ("md", "16x", "8
 .inputWrapper {
   position: relative;
   flex: 1;
+  display: flex;
   min-width: 0;
```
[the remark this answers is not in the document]
atakanbayrak: @motif-ui/core-team  I am not sure about this comment, I looked up for 2 cases and nothing chases visually but there will be changes which will cause a issue for us in advance. What's your opinion ?

src/lib/components/Motif/InputText/InputText.module.scss
aktasmehmet: why do you put block here? It is not necessary at all?

src/lib/components/Motif/InputText/InputText.module.scss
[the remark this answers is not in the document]
aktasmehmet: I think this is invalid. You can resolve this.

aktasmehmet (changes requested):

aktasmehmet (approved):

PR #116, opened 2026-07-23
fix(DateRangePicker, InputDateRange): firstDayOfWeek support

atakanbayrak:
## Description

Since both `DatePicker` and `DateRangePicker` share the same `DaySelector` component, the context was previously used only for `firstDayOfWeek`, while other properties such as `size` and `locale` were already passed via props. To keep the API consistent and remove this unnecessary context dependency, `firstDayOfWeek` is now passed through props as well.

Additionally, managing `firstDayOfWeek` through context caused `DatePickerContext` and `DateRangePickerContext` to influence the shared `DaySelector` independently, creating an implicit coupling between two otherwise unrelated contexts. Passing `firstDayOfWeek` as a prop eliminates this interaction and makes the data flow explicit and predictable.

As part of this change, `firstDayOfWeek` prop support has been added to both `DateRangePicker` and `InputDateRange`. Related stories and test files have been updated, and the behavior has been verified with different locale configurations.

## Type of Change

- [ ] 🆕 New Component
- [x] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Screenshots / Preview

Screenshot 2026-07-23 at 16 41 49
Screenshot 2026-07-23 at 16 41 33

## Checklist

## How to Test

```code
      <DatePicker firstDayOfWeek={3} />
      <DateRangePicker firstDayOfWeek={3} />
      <InputDate firstDayOfWeek={2} />
      <InputDateRange firstDayOfWeek={2} />
```

### DIFF
--- a/src/lib/components/DatePicker/components/DateSelector/Day/DaySelector.tsx
+++ b/src/lib/components/DatePicker/components/DateSelector/Day/DaySelector.tsx
@@ -5,12 +5,14 @@ import styles from "../../../DatePicker.module.scss";
 import { DayProps } from "@/components/DatePicker/components/DateSelector/Day/Day";
 import { Size4SM } from "../../../../../types";
 import { DatePickerLocale } from "@/components/DatePicker/types";
+import { DaysOfWeek } from "@/components/Motif/Pickers/types";
 import Header from "@/components/DatePicker/components/Header/Header";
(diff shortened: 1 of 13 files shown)


aktasmehmet (approved):
🚀 🚀 🚀

PR #120, opened 2026-07-27
fix(InputDateRange): datepicker clipping in overflow-hidden containers

hatesalp:
## Description
- The date range picker was being clipped or hidden when InputDateRange was placed inside a container with overflow: hidden (e.g. Card, Form).
- Introduced usePickerPortal hook that mounts the picker directly to document.body and determines the correct alignment (bottomLeft, bottomRight, topLeft, topRight) based on the available viewport space around the anchor element. Positioning itself is delegated to the existing usePopoverPosition hook rather than reimplementing the logic.
- Updated tests to query picker date buttons from document.body instead of the component container, reflecting the new portal-based rendering.
- Added unit tests for usePickerPortal alignment logic.

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [x] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Screenshots / Preview

**BEFORE:**

{49515FE7-A40E-4ED5-BDC1-E96A4849DCEF}
{932D6C61-543D-4775-9BD7-A01163B4580C}
image

**AFTER:**

{08CB7067-0D82-4997-B44A-70C56F33293F}
{274D01F4-2C5A-4B26-85F4-DC00A068379E}
{9200666C-9561-4189-87DF-C9BEC8A8471C}
{200874CD-D46F-4F88-903A-29CB5FB44461}

## Checklist

## How to Test
 ```
        <Card title="" subtitle="" elevated outlined>
          <Form onSubmit={() => {}}>
            <Grid fluid>
              <Grid.Row>
                <Grid.Col md={3}>
                  <Form.Field name="f1" label="Alan 1 (sol)">
                    <InputDateRange size="sm" />
                  </Form.Field>
                </Grid.Col>

                <Grid.Col md={3}>
                  <Form.Field name="f2" label="Alan 2">
                    <InputDateRange size="sm" />
                  </Form.Field>
                </Grid.Col>

                <Grid.Col md={3}>
                  <Form.Field name="f3" label="Alan 3">
                    <InputDateRange size="sm" />
                  </Form.Field>
                </Grid.Col>

                <Grid.Col md={3}>
                  <Form.Field name="f4" label="Alan 4 (sağ köşe)">
                    <InputDateRange size="sm" />
                  </Form.Field>
                </Grid.Col>
              </Grid.Row>
            </Grid>
          </Form>
        </Card>
```

### CODE REVIEW

File: src/lib/components/InputDateRange/usePickerPortal.ts [hunk 1]
```diff
@@ -0,0 +1,54 @@
...
+  const [alignment, setAlignment] = useState<PickerAlignment>("bottomLeft");
+  const { startShowing, startHiding, attached, positionStyle } = usePopoverPosition(anchorRef, pickerRef, alignment, 0);
+
+  // usePopoverPosition sets --caret-top/--caret-left directly on the DOM node after its
+  // own useLayoutEffect. We clear them in the next layoutEffect (hooks run in order).
+  useLayoutEffect(() => {
+    if (!pickerRef.current) return;
+    pickerRef.current.style.removeProperty("--caret-top");
+    pickerRef.current.style.removeProperty("--caret-left");
+  });
```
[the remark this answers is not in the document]
aktasmehmet: I was gonna say the same thing. Relying on the order makes me nerveous

src/lib/components/InputDateRange/usePickerPortal.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  useEffect(() => {
+    if (visible) startShowing();
+    else startHiding(true);
```
aktasmehmet:
```suggestion
    visible ? startShowing() : startHiding(true);
```

src/lib/components/InputDateRange/usePickerPortal.ts
```diff
@@ -0,0 +1,54 @@
+"use client";
+
+import { useCallback, useEffect, useLayoutEffect, useMemo, useState } from "react";
+import type { RefObject } from "react";
+import { usePopoverPosition } from "@/components/Popover/hooks/usePopoverPosition";
+
+type PickerAlignment = "bottomLeft" | "bottomRight" | "topLeft" | "topRight";
```
aktasmehmet: Instead of creating a new type, that would be clearer to change the position prop type of usePopoverPosition into a seperate type and also use it here.

src/lib/components/InputDateRange/InputDateRange.tsx
```diff
@@ -95,22 +97,24 @@ const InputDateRange = (p: PropsWithRef<InputDateRangeProps, HTMLDivElement>) =>
     !validateRange(itemValue) && clearDateValues();
   }, [clearDateValues, hide, itemValue]);
 
-  const innerRef = useOutsideClick<HTMLDivElement>(outsideClickHandler);
+  const pickerRef = useRef<HTMLDivElement>(null);
+  const pickerExcludedRefs = useMemo(() => [pickerRef], []);
+  const innerRef = useOutsideClick<HTMLDivElement>(outsideClickHandler, pickerExcludedRefs);
```
aktasmehmet:
```suggestion
  const innerRef = useOutsideClick<HTMLDivElement>(outsideClickHandler, [pickerRef]);
```

src/lib/components/InputDateRange/InputDateRange.test.tsx
```diff
@@ -14,6 +25,21 @@ describe("InputDateRange", () => {
...
+  beforeEach(() => {
+    Object.defineProperty(window, "innerWidth", { value: 1024, writable: true });
+    Object.defineProperty(window, "innerHeight", { value: 768, writable: true });
+  });
+
+  afterEach(() => {
+    mockedPosition.mockClear();
+    mockStartShowing.mockClear();
+    mockStartHiding.mockClear();
+  });
```
aktasmehmet: I dont think those should apply in every test. So, you may think yo move these into relevant tests instead of using these methods.

src/lib/components/InputDateRange/InputDateRange.test.tsx
```diff
@@ -14,6 +25,21 @@ describe("InputDateRange", () => {
   const testDateArr = [new Date(2025, 4, 12), new Date(2025, 4, 21)];
   const placeholder = "__ / __ / ____"; // default formata göre "DD/MM/YYYY"
 
+  const mockedPosition = jest.mocked(usePopoverPosition);
+  const mockStartShowing = jest.fn();
+  const mockStartHiding = jest.fn();
```
aktasmehmet: These should be moved into relevant tests. They dont need to be global

src/lib/components/InputDateRange/InputDateRange.test.tsx
```diff
@@ -37,6 +63,37 @@ describe("InputDateRange", () => {
...
+  };
+
+  const mockPositionReturn = (positionStyle: Record<string, number> = { top: 100, left: 50 }) =>
+    mockedPosition.mockReturnValue({
+      startShowing: mockStartShowing,
+      startHiding: mockStartHiding,
+      attached: false,
+      visible: false,
+      positionStyle,
+    });
```
aktasmehmet: You dont need to create a seperated const for that. It only calls a funtion. You can easily move this to the it.each section below

src/lib/components/InputDateRange/usePickerPortal.ts
```diff
@@ -0,0 +1,54 @@
+"use client";
```
aktasmehmet, line 1: I am putting this here. There are some issues when resizing the window or in the scrollable area. Check recordings.

Libs like MUI always tracks and moves the dialog with the anchor. You are free to implement the easiest solution here. I mean, when screen is resized, you may close the popover, etc.

https://github.com/user-attachments/assets/4762bc49-8412-4d3d-8316-864169da69ee
https://github.com/user-attachments/assets/656ae3e5-82df-492a-97d6-755f27165eda

aktasmehmet (commented):
Very clever solution! I've put some minor requests. Using too many hooks (similar hooks) together is not a good solution but you dont have to fix that here. I already created a task to get rid of all of them, inspired by your solution 😉

src/lib/components/InputDateRange/usePickerPortal.ts
hatesalp: Good catch, thanks for pointing it out! I missed this case. I'll fix it and update the PR. :100:

src/lib/components/InputDateRange/usePickerPortal.ts
hatesalp: You're right. I'll update it in the next PR.

src/lib/components/InputDateRange/usePickerPortal.ts
```diff
@@ -0,0 +1,54 @@
...
+
+import { useCallback, useEffect, useLayoutEffect, useMemo, useState } from "react";
+import type { RefObject } from "react";
+import { usePopoverPosition } from "@/components/Popover/hooks/usePopoverPosition";
+
+type PickerAlignment = "bottomLeft" | "bottomRight" | "topLeft" | "topRight";
+
+const PICKER_TOP_OFFSET = 2;
```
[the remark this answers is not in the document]
hatesalp: The value accounts for the anchor's margin-bottom: var(--base-sizing-2x) when calculating the position for the upward-opening picker

src/lib/components/InputDateRange/usePickerPortal.ts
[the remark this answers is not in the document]
hatesalp: I'll update this in the next PR by adding a `showCaret` option to `usePopoverPosition` (defaulting to `true`) and passing `false` here, so the caret variables are never set in the first place. This will also remove the ordering dependency.

src/lib/components/InputDateRange/usePickerPortal.ts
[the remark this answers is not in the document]
aktasmehmet: you have to do it like: 
```
top: `calc(${positionStyle.top} - var(--base-sizing-2x) )`
```

and remove the const

src/lib/components/InputDateRange/usePickerPortal.ts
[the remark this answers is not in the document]
aktasmehmet: what is "next PR"?

src/lib/components/InputDateRange/usePickerPortal.ts
aktasmehmet: I dont understand the next pr and in fact, this is an easy one, 5 minutes. You should do it here

src/lib/components/InputDateRange/InputDateRange.tsx
```diff
@@ -95,21 +97,40 @@ const InputDateRange = (p: PropsWithRef<InputDateRangeProps, HTMLDivElement>) =>
     !validateRange(itemValue) && clearDateValues();
   }, [clearDateValues, hide, itemValue]);
 
-  const innerRef = useOutsideClick<HTMLDivElement>(outsideClickHandler);
+  const pickerRef = useRef<HTMLDivElement>(null);
+  const innerRef = useOutsideClick<HTMLDivElement>(outsideClickHandler, [pickerRef]);
```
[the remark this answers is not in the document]
aktasmehmet: Actually it is right, and its in on me. You may revert here

src/lib/components/InputDateRange/usePickerPortal.ts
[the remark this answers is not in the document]
hatesalp: Ah, sorry. I meant it would be updated in the next commit. It's updated, you can take a look.

src/lib/components/InputDateRange/usePickerPortal.ts
hatesalp: Ah, sorry. I meant it would be updated in the next commit. It's updated now, you can take a look.

src/lib/components/InputDateRange/usePickerPortal.ts
aktasmehmet: This bug continues increasingly :)

Scroll thing is resolve but window resize has 2 bugs:

1 - it is not working if you widen the window
2 - when shrinking it, it disappears but after when clicked outside, it appears strangely.


https://github.com/user-attachments/assets/b3fecb0d-b0b7-4ab5-acb9-fae7c366b84c

src/lib/components/InputDateRange/usePickerPortal.ts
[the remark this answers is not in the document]
aktasmehmet: got it, you tried to say next commit/push 😉

aktasmehmet (changes requested):

src/lib/components/InputDateRange/usePickerPortal.ts
hatesalp: It's fixed now. Could you please check it again? 🙏

src/lib/components/Popover/hooks/usePopoverPosition.tsx
```diff
@@ -1,11 +1,14 @@
 import { RefObject, useLayoutEffect, useCallback, useState, CSSProperties, useRef } from "react";
 import useDomReady from "../../../hooks/useDomReady";
 
+export type PopoverPosition = "top" | "topLeft" | "topRight" | "bottom" | "bottomLeft" | "bottomRight" | "right" | "left";
```
aktasmehmet: Since this is now used in picker, it is not directly related with popover. So, we should move this into /src/lib/types.ts and rename as "OverlayPosition"

src/lib/components/Popover/hooks/usePopoverPosition.tsx
```diff
@@ -71,14 +74,17 @@ export const usePopoverPosition = (
       const finalVisualTop = visualTop + diffY;
       const CARET_OFFSET = 6;
 
-      if (position === "left" || position === "right") {
-        const caretTop = anchorCenterY - finalVisualTop - CARET_OFFSET;
-        itemRef.current.style.setProperty("--caret-top", `${caretTop}px`);
-      } else {
-        const caretLeft = anchorCenterX - finalVisualLeft - CARET_OFFSET;
-        itemRef.current.style.setProperty("--caret-left", `${caretLeft}px`);
+      if (showCaret) {
```
aktasmehmet: Since --caret-top and caret-left only effects Popover, this has nothing to do with the picker. So I dont see a need here. Why dont you revert all **showCaret** stuff in this file?

src/lib/components/InputDateRange/usePickerPortal.ts
```diff
@@ -0,0 +1,100 @@
...
+) => {
+  const [alignment, setAlignment] = useState<PopoverPosition>("bottomLeft");
+  const { startShowing, startHiding, attached, positionStyle } = usePopoverPosition(anchorRef, pickerRef, alignment, 0, false);
+
+  useEffect(() => {
+    if (visible) {
+      startShowing();
+    } else {
+      startHiding(true);
+      anchorRef.current?.querySelector("input")?.blur();
```
aktasmehmet, line 23: is this necessary? Did you put that blur here because of an issue? If not, what is the problem with keeping the focus on the input when the picker is being hidden?

src/lib/components/InputDateRange/InputDateRange.test.tsx
```diff
@@ -266,9 +299,57 @@ describe("InputDateRange", () => {
...
+
+  it("should close the datepicker when the window is resized", async () => {
+    const { getDateRangeInput, getPickerContainer } = renderExt(<InputDateRange />);
+
+    await user.click(getDateRangeInput());
+    expect(getPickerContainer()).toBeInTheDocument();
+
+    await act(() => window.dispatchEvent(new Event("resize")));
+    expect(getPickerContainer()).not.toBeInTheDocument();
   });
```
aktasmehmet, line 353: you should add another test here like "should focus inside the picker when picker is opened and then tab key is pressed"

src/lib/components/InputDateRange/InputDateRange.test.tsx
```diff
@@ -266,9 +299,57 @@ describe("InputDateRange", () => {
...
-    expect(container.firstElementChild?.getElementsByClassName("weekDays")[0].firstElementChild?.textContent).toBe("We");
+    expect(document.getElementsByClassName("weekDays")[0].firstElementChild?.textContent).toBe("We");
+  });
+
+  const alignmentCases = [
+    { label: "top-left region of the viewport", anchorRect: {}, expected: "bottomLeft" },
+    { label: "near the right edge", anchorRect: { left: 800, right: 1000 }, expected: "bottomRight" },
+    { label: "near the bottom edge", anchorRect: { top: 600, bottom: 650 }, expected: "topLeft" },
+    { label: "bottom-right corner", anchorRect: { top: 600, bottom: 650, left: 800, right: 1000 }, expected: "topRight" },
+  ];
```
aktasmehmet: can you move this part to the top?

aktasmehmet (changes requested):

src/lib/components/InputDateRange/usePickerPortal.ts
hatesalp: I added .blur() on close because without it, hiding a focused element inside the portal would return focus to the input, triggering onFocus again and causing the picker to reopen immediately.
I'm adding a video showing the behavior with .blur() disabled.

https://github.com/user-attachments/assets/155741dd-896e-4fe8-b0d9-11c3e9a74122

src/lib/components/Popover/hooks/usePopoverPosition.tsx
hatesalp: Because usePickerPortal uses usePopoverPosition, it sets --caret-left and --caret-top as inline styles on the picker element by default. Since the picker doesn't need these variables, showCaret: false prevents unnecessary style updates during position recalculation.

**BEFORE:** 
Image




**AFTER:**
Image

hatesalp: As discussed in the meeting, `showCaret `was removed and the code was updated accordingly.

aktasmehmet (approved):

PR #121, opened 2026-07-28
fix(Table): locale-aware case normalization for filter matching

hatesalp:
## Description

- Replaced toLowerCase() with toLocaleLowerCase(locale) in TableContext so filter comparisons use the active locale.
- Applied this change to both global search and column filters.
- Added normalization (normalize("NFC")), and mapped ı to i to correctly handle Turkish İ/ı characters.

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [x] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Screenshots / Preview
{40FBBCF2-0AB1-424E-B33E-6A1A7FDE45D7}

{53D76062-E4F9-49AC-8A85-EE700D9A0324}

{3FACD5B9-7ACD-4142-B87E-0E1315604742}

## Checklist

## How to Test
```
const columns = [
  { title: "Ad / Name", dataKey: "ad", filter: true },
  { title: "Şehir / City", dataKey: "sehir", filter: true },
  { title: "Unvan / Title", dataKey: "unvan", filter: true },
];

const data = [
  // Türkçe — büyük/küçük karışımı
  { ad: "İBRAHİM ÇELIK", sehir: "İSTANBUL", unvan: "Müdür" },
  { ad: "Şükrü Öztürk", sehir: "ANKARA", unvan: "İşletmen" },
  { ad: "GÜNEŞ YILMAZ", sehir: "İzmir", unvan: "Uzman" },
  { ad: "GÜNEŞ YILMAZ", sehir: "İZMİR", unvan: "Uzman Yardımcısı" },
  { ad: "Çiğdem Arslan", sehir: "BURSA", unvan: "ŞEF" },
  { ad: "Işıl Doğan", sehir: "Eskişehir", unvan: "MÜDÜR YARDIMCISI" },
  // Yabancı — standart Latin, Fransızca, Almanca, İspanyolca
  { ad: "John Smith", sehir: "NEW YORK", unvan: "Manager" },
  { ad: "ÉLODIE MARTIN", sehir: "Paris", unvan: "Directrice" },
  { ad: "Müller, Hans", sehir: "MÜNCHEN", unvan: "Ingenieur" },
  { ad: "García López", sehir: "madrid", unvan: "DIRECTOR" },
  { ad: "Søren Aaberg", sehir: "Copenhagen", unvan: "Analyst" },
];

<Table columns={columns} data={data} filterableTable border="rowBorders" hoverable />

```


src/lib/components/Table/TableContext.tsx
```diff
@@ -24,6 +27,8 @@ export const TableProvider = (props: PropsWithChildren<TableContextProps>) => {
 
   const [currentPage, setCurrentPage] = useState<number>(1);
   const [mainFilterQuery, setMainFilterQuery] = useState<string>("");
+  const caseNormalize = (s: string, locale: string): string =>
```
aktasmehmet: I think you put ı-i match on purpose. But there are missing characters. Besides, you can improve the normalization to handle strange ascii characters by splitting them out. So, this would be a better approach:

```
const search = useCallback(
    (content: string, q: string) => {
      const TURKISH_FOLD_MAP: Record<string, string> = {
        "\u0131": "i", // ı
        "\u00F6": "o", // ö
        "\u00FC": "u", // ü
      };
      const localeLowerNormalized = (s: string) =>
        s
          .toLocaleLowerCase(locale)
          .normalize("NFKD")
          .replace(/\p{Diacritic}/gu, "")
          .replace(/[\u0131\u00F6\u00FC]/g, c => TURKISH_FOLD_MAP[c]);
      return localeLowerNormalized(content).includes(localeLowerNormalized(q));
    },
    [locale],
  ); 
```

I'll also converted all that into a single function, which is more readable in the file. So, you can directly use this in lines like ``!localeLower(contentToBeSearched).includes(localeLower(columnQuery))``

src/lib/components/Table/Table.test.tsx
```diff
@@ -1057,4 +1061,22 @@ describe("Table", () => {
...
+  it("should match uppercase BERLİN when searching berlin with locale=en", async () => {
+    const { getFilterableTableInput } = renderExt(
+      <MotifProvider locale="en">
+        <Table columns={cols} data={data} filterableTable />
+      </MotifProvider>,
+    );
+    await userEvent.type(getFilterableTableInput(), "berlin");
+    expect(screen.getByText("BERLİN")).toBeInTheDocument();
+    expect(screen.queryByText("İSTANBUL")).not.toBeInTheDocument();
+  });
```
aktasmehmet: These tests are waay too spesific. The name does not cover İZMİR for example? Titles should not include spesific search queries. Besides, you can merge those tests together and test the examples in the test section of the description of this PR. 

Please be careful on naming. 

`should match uppercase BERLİN when searching berlin with locale=en` means `TABLE should match uppercase BERLİN when searching berlin with locale=en` which means nothing

You should write tests like:

`should apply locale-aware filtering in both global and column based search inputs`

and the tests may cover 2 locales and some edge case texts like you have in the description

src/lib/components/Table/TableContext.tsx
hatesalp: After looking into this, it seems that `ö` and `ü` don't need to be included in the fold map because ``.replace(/\p{Diacritic}/gu, "")`` already converts them to `o` and `u` after the `NFKD` normalization step. Only `ı` still needs to be mapped explicitly. If you're okay with this approach as well, I'll update it in my next commit. I'm also planning to move this into the string utils in a follow-up task, since the same logic will likely be needed for the Selectbox filter.

aktasmehmet: I just checked, yes you are correct. We can remove u and o and continue with ı-i only

src/lib/components/Table/Table.test.tsx
hatesalp: As we talked about, I updated the test. I used `İ, I, ı, ö, Ö, ü,` and `Ü `to verify the normalization done by foldNormalize. I didn't include a separate lowercase i because all `i` variants normalize to the same value, so it wouldn't add any extra coverage. I also added a comment in the test to explain this.

PR #124, opened 2026-07-29
fix(Toast): duration prop is set as optional

ZehranurC:
## Description

- Toast stories (Primary + GlobalIconWrapper/ToastIcons) couldn't be reliably captured by Chromatic due to a button-triggered flow and a race with the auto-dismiss timer
- Made duration optional; when omitted, no progress bar renders and no auto-dismiss timer is armed
- Primary now renders synchronously/statically; the real useToast() demo moved to a separate Interactive story, excluded from Chromatic via the !test tag
- ToastIcons in GlobalIconWrapper uses a large duration so the progress bar stays visible while removing any dismiss race
- Added unit test coverage for the new optional-duration behavior

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [x] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Screenshots / Preview

image

## Checklist

## How to Test

I disabled the tag in GlobalIconWrapper stories you can review it from the Storybook Chromatic tab; you can also check Toast's new stories under Components.


src/lib/components/Motif/GlobalIconWrapper/GlobalIconWrapper.stories.tsx
```diff
@@ -28,7 +28,7 @@ import InputNumber from "@/components/InputNumber";
 
 const meta: Meta = {
   title: "Chromatic/GlobalIconWrapper",
-  tags: ["!autodocs", "!dev"],
+  // tags: ["!autodocs", "!dev"],
```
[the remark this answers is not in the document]
atakanbayrak: +1

src/lib/components/Toast/Toast.stories.tsx
```diff
@@ -39,7 +39,17 @@ const Toaster = (props: ToastProps) => {
...
 
 export const Primary: Story = {
-  render: args => {
-    return <Toaster {...args} />;
+  args: {
+    id: "toast-primary",
+    position: "topRight",
+    duration: 24 * 60 * 60 * 1000,
+    closable: true,
+    onDismiss: () => {},
```
atakanbayrak: I am not sure is this needed ?

src/lib/components/Toast/Toast.tsx
```diff
@@ -51,7 +51,7 @@ const Toast = (props: PropsWithRef<ToastProps, HTMLDivElement>) => {
       timer.clear();
     };
 
-    if (closable) {
+    if (closable && duration !== undefined) {
```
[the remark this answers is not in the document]
atakanbayrak: +1

src/lib/components/Motif/GlobalIconWrapper/GlobalIconWrapper.stories.tsx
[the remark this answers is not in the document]
aktasmehmet: +1

src/lib/components/Toast/Toast.stories.tsx
```diff
@@ -39,7 +39,17 @@ const Toaster = (props: ToastProps) => {
...
   },
+  render: args => <Toast {...args} />,
+};
+
+export const Interactive: Story = {
+  tags: ["!test"],
```
aktasmehmet, line 53: What is this for?

src/lib/components/Toast/Toast.stories.tsx
```diff
@@ -39,7 +39,17 @@ const Toaster = (props: ToastProps) => {
...
+  render: args => <Toaster {...args} />,
```
aktasmehmet, line 71: You need to add additional code block here to make it more understandable (I know it wasnt there before). Like this, it is impossible for users to understand:

Image

src/lib/components/Toast/Toast.stories.tsx
```diff
@@ -39,7 +39,17 @@ const Toaster = (props: ToastProps) => {
...
+    id: "toast-primary",
+    position: "topRight",
+    duration: 24 * 60 * 60 * 1000,
+    closable: true,
+    onDismiss: () => {},
   },
+  render: args => <Toast {...args} />,
+};
+
+export const Interactive: Story = {
```
aktasmehmet, line 52: You need to update `Toast.mdx` file's How to show section by adding this story there to make that section more interactive and understandable.

src/lib/components/Toast/types.ts
```diff
@@ -12,7 +12,7 @@ export type ToastProps = {
   content: string;
   variant: ToastVariant;
   position: ToastPosition;
-  duration: number;
+  duration?: number;
```
aktasmehmet, line 15: This should be reflected to the storybook. It looks like only number there:

Image

src/lib/components/Toast/Toast.stories.tsx
```diff
@@ -39,7 +39,17 @@ const Toaster = (props: ToastProps) => {
 };
 
 export const Primary: Story = {
```
aktasmehmet, line 41: Why is this has text control? instead of dropdown?

Image

src/lib/components/Toast/Toast.stories.tsx
```diff
@@ -39,7 +39,17 @@ const Toaster = (props: ToastProps) => {
...
-  render: args => {
-    return <Toaster {...args} />;
+  args: {
```
aktasmehmet, line 46: Duration should be given as undefined here and the control should be disabled because duration is not applicable to primary story

Image

src/lib/components/Motif/GlobalIconWrapper/GlobalIconWrapper.stories.tsx
```diff
@@ -31,7 +31,7 @@ import InputTime from "@/components/InputTime/InputTime";
 
 const meta: Meta = {
   title: "Chromatic/GlobalIconWrapper",
```
aktasmehmet, line 33: In toast icon story, toasts disappear when hover. Check the recording 👇 

https://github.com/user-attachments/assets/8a646591-a169-4337-b74e-edfd311f4952

src/lib/components/Toast/Toast.stories.tsx
```diff
@@ -39,7 +39,17 @@ const Toaster = (props: ToastProps) => {
...
+    id: "toast-primary",
+    position: "topRight",
+    duration: 24 * 60 * 60 * 1000,
+    closable: true,
```
aktasmehmet: There is a bug, I am not sure it is related with this prop but... I hope this is not a component related thing :( Please check the recording...

https://github.com/user-attachments/assets/e5373261-8bf0-4f44-a707-ee21b07293d5

src/lib/components/Toast/Toast.test.tsx
```diff
@@ -116,6 +117,23 @@ describe("Toast", () => {
     jest.useRealTimers();
   });
 
+  it("should not auto-dismiss and should not render a progress bar when duration is not given", () => {
+    jest.useFakeTimers();
+
+    const { getByText, queryByTestId } = render(
+      <Toast id="t1" onDismiss={() => {}} content={content} variant="info" position="topRight" closable />,
```
aktasmehmet: is onDismiss necessary here?

src/lib/components/Toast/Toast.tsx
```diff
@@ -42,7 +42,7 @@ const Toast = (props: PropsWithRef<ToastProps, HTMLDivElement>) => {
     }, 300);
   }, [id, position, onDismiss]);
 
-  const timer = useTimeout(handleDismiss, duration);
+  const timer = useTimeout(handleDismiss, duration ?? 0);
```
aktasmehmet: If duration is undefined, then there is no need fr a timer at all. So, all of other change you've done like in line 54 will be unnecessary.

The thing there is that even if you give duration part as 0 to useTimeout, it still creates a timeout object, which is not necessary at all.

I know making that part optional is a challange because of the hook usage, but you'll handle that 💪

aktasmehmet (changes requested):

src/lib/components/Toast/Toast.stories.tsx
ZehranurC: Interactive is only there for manual use (a real useToast() usage demo); since Chromatic doesn't simulate clicks, it can't capture it meaningfully, so it's excluded via !test . the actual visual regression coverage runs through Primary.

src/lib/components/Toast/Toast.stories.tsx
ZehranurC: It was like that before, but it shouldn't be . you're right, I hadn't noticed it, I'm fixing it

src/lib/components/Toast/Toast.stories.tsx
ZehranurC: I've tested this many times. I think it's related to the storybook 'switch' object. As you'll see in the video, there's no problem when I try it as a 'radio'.


https://github.com/user-attachments/assets/4cc16e52-499f-4c0a-aa7e-6530f3055b64

src/lib/components/Toast/Toast.test.tsx
ZehranurC: onDismiss is only required in tests that render Toast directly, without useToast(). It's not needed in the other tests that use Toaster/useToast(), because the hook wires up onDismiss internally. Here it's a direct usage, so TypeScript requires it ,removing it causes a compile error.

aktasmehmet: Can you check why onDismiss is mandatory in the first place? I cant recall but maybe we should make it optional?

src/lib/components/Toast/Toast.stories.tsx
aktasmehmet: I know others are fine, this is why there is certainly an issue here :) You can check on production, there is no issue there on Toast story. This PR has it.

I personally think that when we started to use the Toast itself directly, changing the the prop triggers a bug or something that a useEffect causes. You should check in detail

src/lib/components/Toast/Toast.test.tsx
ZehranurC: You're right, it doesn't need to be mandatory. I've updated 'onDismiss' to be optional.

src/lib/components/Toast/types.ts
```diff
@@ -8,12 +8,12 @@ export type ToastVariant = "secondary" | "error" | "warning" | "info" | "success
...
   id: string;
-  onDismiss: (id: string, position: ToastPosition) => void;
+  onDismiss?: (id: string, position: ToastPosition) => void;
   content: string;
   variant: ToastVariant;
   position: ToastPosition;
-  duration: number;
-  closable: boolean;
+  duration?: number;
+  closable?: boolean;
```
ZehranurC, line 16: I updated this so that `closeable` defaults to `true` when no 'duration' is specified.

src/lib/components/Toast/Toast.tsx
```diff
@@ -16,7 +16,7 @@ const Toast = (props: PropsWithRef<ToastProps, HTMLDivElement>) => {
     title,
     content,
     duration,
-    closable,
+    closable = true,
```
ZehranurC: I updated this so that `closeable` defaults to `true` when no `duration` is specified.

aktasmehmet: I couldnt get the connection with the duration but, anyway, I was expecting it is to be true by default so, it is all good :)

src/lib/components/Toast/Toast.stories.tsx
aktasmehmet: This is still valid I guess.

src/lib/components/Motif/GlobalIconWrapper/GlobalIconWrapper.stories.tsx
```diff
@@ -330,7 +330,6 @@ export const ToastIcons: Story = {
           content="Toast message"
           variant="success"
           position="topRight"
-          duration={3000}
           closable
           onDismiss={() => {}}
```
aktasmehmet: onDismiss needs to be removed I guess?

src/lib/components/Toast/Toast.tsx
aktasmehmet: I genuinely think that you put this to prevent the timer start on line #55.

This is not the right approach because now this component is the one managed by useToast. So, it should not put default things, the hook should decide.

What you can do is to revert this default value set and then put ``duration !== undefined`` to the condition on line #55 and the dependency array. That's it. Pleae do and test it.

src/lib/components/Toast/Toast.test.tsx
```diff
@@ -116,6 +117,21 @@ describe("Toast", () => {
     jest.useRealTimers();
   });
 
+  it("should not auto-dismiss and should not render a progress bar when duration is not given", () => {
```
aktasmehmet:
```suggestion
  it("should not auto-dismiss and render a progress bar when duration is undefined", () => {
```

src/lib/components/Toast/Toast.test.tsx
```diff
@@ -116,6 +117,21 @@ describe("Toast", () => {
...
+    expect(getByText(content)).toBeInTheDocument();
+    expect(queryByTestId("progressBar")).not.toBeInTheDocument();
+
+    act(() => {
+      jest.advanceTimersByTime(10000);
+    });
+    expect(getByText(content)).toBeInTheDocument();
+
+    jest.useRealTimers();
+  });
```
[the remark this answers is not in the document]
aktasmehmet: This is right. You should put another test like `should not start a progress bar, timer and dismiss by mouse enter and exits when duration is undefined`

src/lib/components/Toast/Toast.stories.tsx
```diff
@@ -11,8 +11,13 @@ const meta: Meta<typeof Toast> = {
   argTypes: {
     id: { table: { disable: true } },
     onDismiss: { table: { disable: true } },
-    position: { table: { defaultValue: { summary: "topRight" } }, type: { name: "string", required: false } },
-    duration: { table: { defaultValue: { summary: "3000" } }, type: { name: "number", required: false } },
+    position: {
+      table: { defaultValue: { summary: "topRight" } },
+      type: { name: "string", required: false },
+      control: { type: "select" },
+      options: ["topLeft", "topRight", "top", "bottomLeft", "bottomRight", "bottom"],
```
aktasmehmet: Why do we give this? Storybook should automatically generate the dropdown. Is there an issue with the type?

aktasmehmet: Anyway, I think you need to remove the controller completely because we dont want the use to change this in primary story

src/lib/components/Toast/Toast.stories.tsx
```diff
@@ -39,7 +44,34 @@ const Toaster = (props: ToastProps) => {
 };
 
 export const Primary: Story = {
-  render: args => {
-    return <Toaster {...args} />;
+  args: {
```
aktasmehmet, line 46: you need to move some of the argTypes values into this story. Because when you give it globally on top (for example disabling duration), you also cannot the duration in Interactive story (you can try). Those blockages should only apply to primary, not all stories.

src/lib/components/Toast/Toast.stories.tsx
```diff
@@ -39,7 +44,34 @@ const Toaster = (props: ToastProps) => {
...
+    id: "toast-primary",
+    position: "topRight",
+    duration: 24 * 60 * 60 * 1000,
```
aktasmehmet: duration is also not necessary here?

aktasmehmet (changes requested):

src/lib/components/Motif/GlobalIconWrapper/GlobalIconWrapper.stories.tsx
atakanbayrak: +1

src/lib/components/Toast/Toast.stories.tsx
ZehranurC: Actually, it's not needed anymore, I'll remove it.

src/lib/components/Toast/Toast.stories.tsx
ZehranurC: You're right, I came across an existing issue while looking into this. If closable changes while it's still closing, the pending dismiss gets cancelled and the toast stays stuck . two separate timers were interfering with each other. I split them into two separate useEffects so they no longer depend on each other.

src/lib/components/Toast/Toast.stories.tsx
```diff
@@ -11,7 +11,7 @@ const meta: Meta<typeof Toast> = {
   argTypes: {
     id: { table: { disable: true } },
     onDismiss: { table: { disable: true } },
-    position: { table: { defaultValue: { summary: "topRight" } }, type: { name: "string", required: false } },
+    position: { table: { defaultValue: { summary: "topRight" } } },
     duration: { table: { defaultValue: { summary: "3000" } }, type: { name: "number", required: false } },
     closable: { table: { defaultValue: { summary: "true" } }, type: { name: "boolean", required: false } },
```
aktasmehmet: type and required parts are not necessary here

src/lib/components/Toast/Toast.stories.tsx
aktasmehmet: No you got me wrong. I mean that your comment is valid, it has to be removed :)

src/lib/components/Toast/Toast.stories.tsx
```diff
@@ -39,7 +39,37 @@ const Toaster = (props: ToastProps) => {
...
-  render: args => {
-    return <Toaster {...args} />;
+  argTypes: {
+    duration: { control: false },
+    position: { control: false },
+  },
+  args: {
+    id: "toast-primary",
+    position: "topRight",
+    closable: true,
```
aktasmehmet: isnt ie true by default? you dont need to give it here again

aktasmehmet (approved):

hatesalp (approved):

PR #130, opened 2026-08-04
fix(Icon): inline-flex overflow issue

aktasmehmet:
## Description

Using inline-flex as display prop in icons were wrong and causing overflow issues as expected. This PR fixes it

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [x] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Screenshots / Preview

BEFORE

Screenshot 2026-08-04 at 15 47 07

AFTER

Screenshot 2026-08-04 at 15 47 35

## Checklist

## How to Test

Just use <Icon /> anywhere


hatesalp (approved):
I checked all components that use icons, both inside and outside forms. Everything looks good.

PR #140, opened 2026-08-21
fix: new component AlertModal created 🧩

hatesalp:
## Description
- New `AlertModal` component: portal-based, animated (fade + pop), with theme defaults support
- Split into `AlertModalContent` and `AlertModalActions` sub-components
- Intentionally does not support `children` — designed for short, structured alerts (icon, text, actions only), as agreed with the design team. For richer content, the existing `Modal` component should be used instead.
- **Why `ReactElement` for `icon` and `buttonAction`?**
Consumers pass fully configured JSX elements so they retain complete control over the icon's variant and size, and the button's label, shape, and click behavior — without `AlertModal` restricting those choices. This was a deliberate decision to keep the component flexible while still maintaining a fixed layout structure.

**Content:** `title`, `subtitle`, `icon`, `contentPosition` (`left | center | right`, default `center`)
**Visibility & Interaction:** `open`, `onClose`
**Actions:** `buttonAction`, `buttonsPosition` (`left | center | right | spread | stretch`, default `center`), `enableDivider` (default `true`)
**Appearance:** `size` (`sm | md | lg`, default `md`), `removeBackdrop`, `bordered`, `elevated`

## Type of Change

- [x] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Screenshots / Preview
image
{1C32A119-99CE-4EF4-B00D-F761383F1D10}
{C29321DB-02D0-40F7-9C25-88E7958FAA37}
{4F18C55B-8B7C-4439-829E-72ECA25F28FB}

{8460F719-B902-4ADD-B593-9C2E86072C98}
{750FDCA4-38A0-49E1-B842-65352061666A}
{6707C77E-7009-47AF-BC4C-61CD7BE2776A}
{D33DD804-64F5-4263-9C73-0CBA1BB05578}

## Checklist

## How to Test

```
 <AlertModal
        title="Test Modal Content Title"
        subtitle="Modal Content Text Lorem ipsum dolor sit amet, consectetur adipiscing elit. Proin sed tristique tellus. Morbi vel sodales urna. Duis felis leo, lacinia id imperdiet ac, molestie ut neque. "
        icon={<MotifIcon name="info" size="xxl" variant="primary" />}
        size="md"
        open={open}
        onClose={() => setOpen(false)}
        buttonAction={[
          <Button key="negative" label="Vazgeç" pill shape="outline" onClick={() => setOpen(false)} />,
          <Button key="positive" label="Onayla" pill onClick={() => setOpen(false)} />,
        ]}
        contentPosition="right"
        buttonsPosition="center"
      />
```

```
<AlertModal
        title="Test Modal Content Title"
        subtitle="Modal Content Text Lorem ipsum dolor sit amet, consectetur adipiscing elit. Proin sed tristique tellus. Morbi vel sodales urna. Duis felis leo, lacinia id imperdiet ac, molestie ut neque. "
        icon={<MotifIcon name="info" size="xxl" variant="primary" />}
        size="md"
        open={open}
        onClose={() => setOpen(false)}
        buttonAction={[
          <Button key="negative" label="Vazgeç" pill fluid shape="outline" onClick={() => setOpen(false)} />,
          <Button key="positive" label="Onayla" pill fluid onClick={() => setOpen(false)} />,
        ]}
        contentPosition="right"
        buttonsPosition="center"
      />
```


src/lib/components/AlertModal/components/AlertModalActions.tsx
```diff
@@ -0,0 +1,25 @@
...
+import styles from "../AlertModal.module.scss";
+import { AlertModalButtonsPosition } from "../types";
+
+type Props = {
+  buttonAction?: ReactElement<ButtonProps>[];
+  buttonsPosition?: AlertModalButtonsPosition;
+  enableDivider?: boolean;
+};
+
+const AlertModalActions = ({ buttonAction, buttonsPosition = "center", enableDivider = true }: Props) => {
```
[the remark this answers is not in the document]
hatesalp: It will be updated in the next commit.

src/lib/components/AlertModal/types.ts
```diff
@@ -0,0 +1,25 @@
+import { ReactElement } from "react";
+import { ButtonProps } from "../Button/types";
+import { Size3 } from "../../types";
+
+export type AlertModalContentPosition = "left" | "center" | "right";
+export type AlertModalButtonsPosition = AlertModalContentPosition | "spread" | "stretch";
+
+export type AlertModalProps = {
+  title: string;
+  subtitle?: string;
```
aktasmehmet: This should be renamed as "text" as it is in the task. There is no subtitle at all.

src/lib/components/AlertModal/types.ts
```diff
@@ -0,0 +1,25 @@
...
+  icon?: ReactElement;
```
aktasmehmet: This type should be `IconGlobalType`

src/lib/components/AlertModal/types.ts
```diff
@@ -0,0 +1,25 @@
+import { ReactElement } from "react";
+import { ButtonProps } from "../Button/types";
+import { Size3 } from "../../types";
+
+export type AlertModalContentPosition = "left" | "center" | "right";
+export type AlertModalButtonsPosition = AlertModalContentPosition | "spread" | "stretch";
+
+export type AlertModalProps = {
```
aktasmehmet, line 15: `variant` is missing. (That is stated in the task)

src/lib/components/AlertModal/types.ts
```diff
@@ -0,0 +1,25 @@
...
+  title: string;
+  subtitle?: string;
+  icon?: ReactElement;
+  open?: boolean;
+  onClose?: () => void;
+  buttonAction?: ReactElement<ButtonProps>[];
```
aktasmehmet: This should be divided into positive and negative buttons as stated in the task. There is no need for a button list. This is a component of components and will be as simple as possible. 

Another note is that this should not be ReactElement of Buttons. If we give the freedom of using Button, we cannot control the variant or fluid, etc. for example if right now you give buttons with fluid here, buttonsPositions not working at all.

So this should be an object of icon, label, onClick for positive and negative button props.

src/lib/components/AlertModal/types.ts
```diff
@@ -0,0 +1,25 @@
...
+} & AlertModalDefaultableProps;
+
+export type AlertModalDefaultableProps = {
+  size?: Size3;
```
aktasmehmet: There should be a fluid or something similar to enlarge this to fill the screen. Similar is also available in the modal I guess. That should be added to Size3 here.

aktasmehmet: This should follow the padding rules and should not stick to the very edge of the scren of course.

src/lib/components/AlertModal/types.ts
aktasmehmet: This icon prop may also be moved into defaultables in case to create an app wide AlertModal with a single icon.

src/lib/components/AlertModal/types.ts
```diff
@@ -0,0 +1,25 @@
+import { ReactElement } from "react";
+import { ButtonProps } from "../Button/types";
+import { Size3 } from "../../types";
+
+export type AlertModalContentPosition = "left" | "center" | "right";
+export type AlertModalButtonsPosition = AlertModalContentPosition | "spread" | "stretch";
```
aktasmehmet: These are unnecessary and can be inlined. Besides, here Content position overrides Buttons position which is also unrelated.

src/lib/components/AlertModal/AlertModal.module.scss
```diff
@@ -0,0 +1,135 @@
...
+  }
+}
+
+.textContent {
+  display: flex;
+  flex-direction: column;
+  gap: var(--base-sizing-8x);
+}
+
+.title {
```
aktasmehmet: There is a vertical alignment issue with title and the icon

Image

src/lib/components/AlertModal/types.ts
aktasmehmet: Image

src/lib/components/AlertModal/types.ts
```diff
@@ -0,0 +1,25 @@
...
+  subtitle?: string;
+  icon?: ReactElement;
+  open?: boolean;
+  onClose?: () => void;
+  buttonAction?: ReactElement<ButtonProps>[];
+} & AlertModalDefaultableProps;
+
+export type AlertModalDefaultableProps = {
+  size?: Size3;
+  contentPosition?: AlertModalContentPosition;
```
aktasmehmet, line 27: When contentPosition is right, the icon also positions itself to the right side of the title. Can you discuss this with the design team. They might ask to keep the icon on the left side of the title even if the title is aligned on the right

src/lib/components/AlertModal/components/AlertModalActions.tsx
aktasmehmet: enableDivider should not be true by default

src/lib/components/AlertModal/components/AlertModalActions.tsx
```diff
@@ -0,0 +1,25 @@
...
+  if (!buttonAction?.length) return null;
+  const classNames = `${styles.actions} ${styles[buttonsPosition]}`;
+
+  return (
+    <>
+      {enableDivider && <Divider className={styles.divider} />}
```
aktasmehmet: it is not necessary to couple this to Divider component. Just use a basic border-top? Besides, this is 2px and in figma it is 1px. That should also be fixed

src/lib/components/AlertModal/AlertModal.module.scss
```diff
@@ -0,0 +1,135 @@
...
+  color: var(--theme-color-text-base-title);
+  @include font("heading", "heading5");
+}
+
+.subtitle {
+  @include font("body", "md");
```
aktasmehmet: There is no color attribute here (Figma has it)

src/lib/components/AlertModal/types.ts
```diff
@@ -0,0 +1,25 @@
...
+  buttonsPosition?: AlertModalButtonsPosition;
+  enableDivider?: boolean;
+  removeBackdrop?: boolean;
+  bordered?: boolean;
+  elevated?: boolean;
+};
```
aktasmehmet, line 35: There should be a "solid" prop here. That was in the modal itself I guess (with colored background etc). It was stated in the task as well.

Besides, I guess this is not in the Figma as well. So, that should also be told to the designers to update there.

src/lib/components/AlertModal/AlertModal.tsx
```diff
@@ -0,0 +1,84 @@
...
+    ref,
+    style,
+    className,
+  } = usePropsWithThemeDefaults("AlertModal", props);
+
+  const domReady = useDomReady();
+  const [visible, setVisible] = useState(open);
+  const [attached, setAttached] = useState(open);
+
+  const handleCloseAnimation = useCallback(() => {
```
aktasmehmet: Why dont you useToggle here? It has everything you've done here and in fact intented to use in this type of scenarios.

src/lib/components/AlertModal/components/AlertModalActions.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { AlertModalButtonsPosition } from "../types";
+
+type Props = {
+  buttonAction?: ReactElement<ButtonProps>[];
+  buttonsPosition?: AlertModalButtonsPosition;
+  enableDivider?: boolean;
+};
+
+const AlertModalActions = ({ buttonAction, buttonsPosition = "center", enableDivider = true }: Props) => {
+  if (!buttonAction?.length) return null;
```
aktasmehmet: This is not a pattern that we use. If render of a component depends on a prop, that should be handles outside of the component, in the parent.

src/lib/components/AlertModal/components/AlertModalContent.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+type Props = {
+  title: string;
+  subtitle?: string;
+  contentPosition?: AlertModalContentPosition;
+  icon?: ReactElement;
+};
+
+const AlertModalContent = ({ title, subtitle, contentPosition = "center", icon }: Props) => {
+  const bodyClasses = `${styles.content} ${styles[contentPosition]}`;
```
aktasmehmet: This should be renamed as classNames. It is like that everywhere unless there are other class names that create confusion.

src/lib/components/AlertModal/components/AlertModalActions.tsx
```diff
@@ -0,0 +1,25 @@
...
+  const classNames = `${styles.actions} ${styles[buttonsPosition]}`;
```
aktasmehmet: Here and also in AlertModalContent, us sanitization methods. For example in the future if someone removes the default value of "buttonsPosition" here, this class will be set as `undefined`.

That is the main point that we've created that function.

src/lib/components/AlertModal/components/AlertModalContent.tsx
aktasmehmet: Giving class names like contentPosition or buttonsPosition may lead to some issues. Thing about that: here, the outcome will be "content right", "content center", etc. But similar `right` and `center` class names will also be set in AlertModalActions. How to distinguish them?

This is the reason that we follow BEM here. So, update the class names accordingly, like "buttons_right", etc.

In fact, by doing that, you can also move those classes to the very top (root element) of the component and remove from these sub components to make them more readable. Actually that was our pattern to follow, put the class names to the top as much as possible.

src/lib/components/AlertModal/AlertModal.module.scss
```diff
@@ -0,0 +1,135 @@
...
+
+  &.show {
+    @include fade-in;
+
+    .alertModal {
+      @include pop-in;
+    }
+  }
+
+  &.noBackdrop {
```
aktasmehmet: This is not a nice class name. In fact, the root element has backdrop class which is not nice also. I'd rename the root elements class name as "alertModalContainer" and put "backdrop" class if there is a backdrop and remove noBackdrop

src/lib/components/AlertModal/AlertModal.module.scss
```diff
@@ -0,0 +1,135 @@
...
+
+  .alertModal {
+    display: flex;
+    flex-direction: column;
+    background: var(--theme-color-surface-base-bg-light);
+    color: var(--theme-color-text-base-title);
+    padding: var(--base-sizing-8x);
+    border-radius: var(--theme-sizing-radius-default);
+    max-width: calc(100% - 4rem);
+    @include shadow(4);
```
aktasmehmet: This is wrong and make the component elevated even if elevated is false.

Image

src/lib/components/AlertModal/AlertModal.module.scss
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  .alertModal {
+    display: flex;
+    flex-direction: column;
+    background: var(--theme-color-surface-base-bg-light);
+    color: var(--theme-color-text-base-title);
+    padding: var(--base-sizing-8x);
+    border-radius: var(--theme-sizing-radius-default);
+    max-width: calc(100% - 4rem);
```
aktasmehmet: is this necessary? I mean, you already have a padding in the root which prevents the content to stick to the edges, right? You just may need to increase that padding a little bit (to make shadows look better)

and remove the max width-here of course 😉

src/lib/components/AlertModal/AlertModal.module.scss
```diff
@@ -0,0 +1,135 @@
...
+  color: var(--theme-color-text-base-title);
+  @include font("heading", "heading5");
+}
+
+.subtitle {
+  @include font("body", "md");
+}
+
+.divider {
+  margin: 0 !important;
```
aktasmehmet: you can find a way to get rid of this important 😉

src/lib/components/AlertModal/AlertModal.module.scss
```diff
@@ -0,0 +1,135 @@
...
+    background-color: transparent;
+  }
+
+  &.elevated .alertModal {
+    @include shadow(6);
+  }
+
+  &.bordered .alertModal {
+    border: solid var(--base-sizing-1x) var(--theme-color-border-light-default);
+  }
```
aktasmehmet: Those classes inside the root can be move outside of the root. This is our pattern: keep the spesifity as low as possible.

src/lib/components/AlertModal/AlertModal.module.scss
```diff
@@ -0,0 +1,135 @@
...
+  &.stretch {
+    & > * {
+      flex: 1;
+    }
+  }
+
+  @each $position, $val in $button-positions {
+    &.#{$position} {
+      justify-content: $val;
+      align-items: $val;
```
aktasmehmet: align-items serves nothing here.

src/lib/components/AlertModal/AlertModal.module.scss
```diff
@@ -0,0 +1,135 @@
...
+    justify-content: space-between;
+  }
+
+  &.stretch {
+    & > * {
+      flex: 1;
+    }
+  }
+
+  @each $position, $val in $button-positions {
```
aktasmehmet: when align-items is removed there, this button-positions will serve for only one attr and there is only 3 items. So, please remove that and put them as hardcoded here. No need for overcomplexity.

src/lib/components/AlertModal/AlertModal.stories.tsx
```diff
@@ -0,0 +1,138 @@
...
+import useToggle from "../../hooks/useToggle";
+import { formatStoryTransform } from "../../../utils/docUtils";
+
+const meta: Meta<typeof AlertModal> = {
+  title: "Components/AlertModal",
+  component: AlertModal,
+  argTypes: {
+    open: { control: false },
+    buttonAction: { control: false },
+    icon: { control: false },
```
aktasmehmet: why is this control false? We can give whatever button we want, as in other components?

src/lib/components/AlertModal/AlertModal.stories.tsx
```diff
@@ -0,0 +1,138 @@
...
+import MotifIcon from "../Motif/Icon/MotifIcon";
+import useToggle from "../../hooks/useToggle";
+import { formatStoryTransform } from "../../../utils/docUtils";
+
+const meta: Meta<typeof AlertModal> = {
+  title: "Components/AlertModal",
+  component: AlertModal,
+  argTypes: {
+    open: { control: false },
+    buttonAction: { control: false },
```
aktasmehmet: When you update buttonAction (positive and negative) those should not be set as control false

src/lib/components/AlertModal/AlertModal.stories.tsx
```diff
@@ -0,0 +1,138 @@
...
+    />
+  </>
+);`;
+        }),
+      },
+    },
+  },
+};
+
+const FluidAlertModalComponent = (props: AlertModalProps) => {
```
aktasmehmet: There is no need for this story. When you update buttonsPositions with fluid value, this scenarion will be available in the primary story.

src/lib/components/AlertModal/AlertModal.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+    getBackdrop,
+  };
+};
+
+describe("AlertModal", () => {
+  it("should be rendered with only required props and should have default prop values stated here", () => {
+    const { container, getBackdrop } = renderExt(<AlertModal title="Alert Modal Title" open />);
+    expect(container).toMatchSnapshot();
+
+    // size: md
```
aktasmehmet: There is only one default prop control here. Where are the rest?

src/lib/components/AlertModal/AlertModal.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+  it("should render with border when bordered prop is true", () => {
+    const { getBackdrop } = renderExt(<AlertModal title="Alert Modal Title" open bordered />);
+    expect(getBackdrop()).toHaveClass("bordered");
+  });
+
+  it("should render as elevated when elevated prop is true", () => {
+    const { getBackdrop } = renderExt(<AlertModal title="Alert Modal Title" open elevated />);
+    expect(getBackdrop()).toHaveClass("elevated");
+  });
+});
```
aktasmehmet, line 222: Please double check here for the missing tests for each prop. I've seen 10 tests here out of 13 props.

aktasmehmet (changes requested):

src/lib/components/AlertModal/types.ts
hatesalp: I talked to the design team. They decided to keep only the `left` and `center` options.

src/lib/components/AlertModal/AlertModal.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  it("should render the subtitle given in the subtitle prop", () => {
+    renderExt(<AlertModal title="Alert Modal Title" open subtitle="Alert Subtitle" />);
+    expect(screen.getByText("Alert Subtitle")).toBeInTheDocument();
+  });
+
+  it("should render the icon given in the icon prop", () => {
+    renderExt(<AlertModal title="Alert Modal Title" open icon={<MotifIcon name="home" />} />);
+    expect(screen.getByText("home")).toBeInTheDocument();
+  });
```
atakanbayrak, line 76: In these 3 tests; title, subtitle and icon was tested from screen that if it is in DOM or not. Specific catch would be better if component have them in the correct places.

src/lib/components/Modal/Modal.test.tsx
```diff
@@ -138,7 +138,7 @@ describe("Modal", () => {
   });
 
   it("should be rendered with the given size in size prop", () => {
-    const sizes: Size3[] = ["sm", "md", "lg"];
+    const sizes: Size4LG[] = ["sm", "md", "lg", "xl"];
```
hatesalp: I'll remove the `xl` size

src/lib/components/Motif/Icon/MotifIconButton.test.tsx
```diff
@@ -24,7 +24,7 @@ describe("MotifIconButton", () => {
   });
 
   it("should render with the size given in the size prop", () => {
-    const sizes: Size7[] = ["xxs", "xs", "sm", "md", "lg", "xl", "xxl"];
+    const sizes: Size7[] = ["xxs", "xs", "sm", "md", "lg", "xl", "xl2"];
```
hatesalp: I'll add the xl3 size.

src/lib/components/AlertModal/AlertModal.stories.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const { open, ...otherProps } = props;
+  const { visible, show, hide } = useToggle(open);
+
+  return (
+    <>
+      <Button label="Open Modal" onClick={show} />
+      <AlertModal
+        open={visible}
+        onClose={hide}
+        icon={<MotifIcon name="info" variant="primary" size="xxl" />}
```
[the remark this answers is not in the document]
hatesalp: Updated in the latest commit.

hatesalp:
## AlertModal

* Supports `title`, `text`, `icon`, `variant`, `size`, `contentPosition`, `buttonsPosition`, `actionButton`, `alternateButton`, `enableDivider`, `removeBackdrop`, `bordered`, `elevated`, `open`, and `onClose`.

## useToggle Refactor

* Changed `toggleTime` from a single parameter to an options object with `showTime` and `hideTime`.
* `hide()` now sets `visible` to `false` immediately and keeps the component in the DOM until `hideTime` ends.
* This allows CSS exit transitions to start immediately and moves the animation logic into `useToggle`.
* Removed the manual `handleCloseWithAnimation` and `setTimeout` logic from `Modal` and `AlertModal`.

## Size Tokens

* Renamed `xxl` to `xl2`.
* Added `xl3` size to `Icon`.

## Breaking Changes

* `useToggle(initialVisible, toggleTime)` → `useToggle(initialVisible, { showTime?, hideTime? })`

src/lib/components/Alert/Alert.tsx
```diff
@@ -19,12 +19,12 @@ const Alert = (props: PropsWithRefAndChildren<AlertProps, HTMLDivElement>) => {
     style,
   } = usePropsWithThemeDefaults("Alert", props);
 
-  const { visible, hide, toggleState } = useToggle(true, 300);
+  const { visible, hide, toggleState } = useToggle(true, { hideTime: 300 });
```
hatesalp: It will be updated to `useToggle(true, { showTime: 300, hideTime: 300 })`

src/lib/components/AlertModal/types.ts
aktasmehmet: we decided not to add that

src/lib/components/AlertModal/types.ts
aktasmehmet: Thx, looks fine. "icon" prop is missing in the buttons.

src/lib/components/AlertModal/types.ts
aktasmehmet: objection! :) Even if it is not a good looking option, I cannot accept a prop with value "center" and "left". We should also add right here no matter what. The first thing comes to my mind is RTL support 😉 

So, please discuss this again and put back that option.

src/lib/components/AlertModal/components/AlertModalActions.tsx
aktasmehmet: You've transformed this into 

``isVisible &&``

in the component. So my suggestion is to move this logic out of AlertModalActions and put something like this to the parent:

``isVisible && <AlertModalActions />``

This would prevent unnecessary component mount.

src/lib/components/AlertModal/components/AlertModalContent.tsx
aktasmehmet: This time you've created a className prop and pass the same class name from the parent :) You dont need that. You can put those necessary class names to the root of the component and handle everything in the scss

src/lib/components/AlertModal/components/AlertModalActions.tsx
aktasmehmet: Same is also applicable here:

https://github.com/motif-ui/motifreact/pull/140#discussion_r3965625576

src/lib/components/AlertModal/AlertModal.module.scss
aktasmehmet: They are still in the root. This is not resolved.

src/lib/components/AlertModal/AlertModal.test.tsx
aktasmehmet: we dont put default false/undefined props here, ever. Never did that before:

``
// bordered: false
    expect(getBackdrop()).not.toHaveClass("bordered");
    // elevated: false
    expect(getBackdrop()).not.toHaveClass("elevated");
``

Please remove those unnecessary checks there.

aktasmehmet (changes requested):
There are still some things to fix. I've resolved most of them but I decided to wait until remainings are resolved and check the component from scratch.

PS. I didnt look in deep but I am not a fan of dividing the toggle time into 2 sections like start and finish. This is a very detailed functionality which will break and make a need for refactoring for lots of components, unnecessarily.

src/lib/components/AlertModal/AlertModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+  ]);
+
+  const contentClassNames = sanitizeModuleClasses(styles, "content", `content_${contentPosition}`);
+  const actionsClassNames = sanitizeModuleClasses(styles, "actions", `actions_${buttonsPosition}`, enableDivider && "withDivider");
+
+  return (
+    attached &&
+    domReady &&
+    createPortal(
+      <div data-testid="alertModalBackdrop" className={classNames} style={style} ref={ref}>
```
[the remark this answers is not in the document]
hatesalp: We’re skipping the a11y items for now.

src/lib/components/AlertModal/AlertModal.test.tsx
```diff
@@ -0,0 +1,189 @@
...
+    getBackdrop,
+    getModalActions,
+    getModalContent,
+  };
+};
+
+describe("AlertModal", () => {
+  it("should be rendered with only required props and should have default prop values stated here", () => {
+    const { container, getBackdrop, getModalActions } = renderExt(
+      <AlertModal title="Alert Modal Title" open actionButton={{ text: "Confirm", onClick: jest.fn() }} />,
```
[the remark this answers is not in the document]
hatesalp: updated.

src/lib/components/AlertModal/components/AlertModalActions.tsx
aktasmehmet: See the change. Great, but, `isVisible` is not a good anming there, because now you are in AlertModal and that name refers to that component. So, you may delete that const and just use `` actionButton || alternateButton `` instead?

src/lib/components/AlertModal/components/AlertModalContent.tsx
aktasmehmet: You can still consider some class names to be pushed up to the parent

src/lib/components/AlertModal/AlertModal.module.scss
aktasmehmet: Example: `backdrop` and `show` classes

src/lib/components/AlertModal/AlertModal.test.tsx
aktasmehmet: `buttonPosition` and `variant`default values are not tested still

src/lib/components/AlertModal/components/AlertModalContent.tsx
```diff
@@ -0,0 +1,30 @@
+import { memo } from "react";
+import styles from "../AlertModal.module.scss";
+import { IconGlobalType, Variant } from "../../../types";
+import GlobalIconWrapper from "../../Motif/GlobalIconWrapper/GlobalIconWrapper";
+import { sanitizeModuleClasses } from "src/utils/cssUtils";
```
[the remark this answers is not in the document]
hatesalp: It will be removed in the next update.

src/lib/components/AlertModal/AlertModal.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+    const { getBackdrop } = renderExt(<AlertModal title="Alert Modal Title" open bordered />);
+    expect(getBackdrop()).toHaveClass("bordered");
+  });
+
+  it("should render as elevated when elevated prop is true", () => {
+    const { getBackdrop } = renderExt(<AlertModal title="Alert Modal Title" open elevated />);
+    expect(getBackdrop()).toHaveClass("elevated");
+  });
+
+  it("should apply variant class to icon when variant is provided", () => {
```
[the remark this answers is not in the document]
hatesalp: It will be added in the next update.

src/lib/components/AlertModal/AlertModal.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+  });
+
+  it("should render the modal when open is true", () => {
+    const { rerender, queryByTestId, getBackdrop } = renderExt(<AlertModal title="Alert Modal Title" open={false} />);
+    expect(queryByTestId("alertModalBackdrop")).not.toBeInTheDocument();
+    rerender(<AlertModal title="Alert Modal Title" open />);
+    expect(getBackdrop()).toBeInTheDocument();
+  });
+
+  it("should render the title given in the title prop", () => {
```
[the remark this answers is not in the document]
hatesalp: As with the other components, it should stay this way.

src/lib/components/AlertModal/AlertModal.test.tsx
hatesalp: The `variant` prop is not passed to the CSS, so the variant class is not added to the root class. Although the default value is `primary`, it is still used for the icon and button colors. I added a separate test for this. I'm not sure if we also need to test this case here.

hatesalp: I added it specifically for the icon and button colors.

src/lib/components/AlertModal/AlertModal.test.tsx
```diff
@@ -0,0 +1,200 @@
...
+  it("should render alternate button and call handler when clicked", () => {
+    const handleCancel = jest.fn();
+    renderExt(<AlertModal title="Alert Modal Title" open alternateButton={{ text: "Cancel", onClick: handleCancel }} />);
+    fireEvent.click(screen.getByText("Cancel"));
+    expect(handleCancel).toHaveBeenCalledTimes(1);
+  });
+
+  it("should apply the given contentPosition class", () => {
+    const contentPosition: AlertModalContentPosition[] = ["left", "center", "right"];
+    contentPosition.forEach(contentPosition => {
```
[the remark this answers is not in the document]
hatesalp: It's fine to keep it this way.

src/lib/components/AlertModal/AlertModal.test.tsx
```diff
@@ -0,0 +1,200 @@
...
+    getModalContent,
+  };
+};
+
+describe("AlertModal", () => {
+  it("should be rendered with only required props and should have default prop values stated here", () => {
+    const { container, getBackdrop, getModalActions } = renderExt(
+      <AlertModal title="Alert Modal Title" open icon={<MotifIcon name="home" />} actionButton={{ text: "Confirm", onClick: jest.fn() }} />,
+    );
+    expect(container).toMatchSnapshot();
```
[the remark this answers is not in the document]
hatesalp: It should stay this way.

src/lib/components/AlertModal/AlertModal.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+    getBackdrop,
+    getModalActions,
+    getModalContent,
+  };
+};
+
+describe("AlertModal", () => {
+  it("should be rendered with only required props and should have default prop values stated here", () => {
+    const { container, getBackdrop } = renderExt(<AlertModal title="Alert Modal Title" open />);
+    expect(container).toMatchSnapshot();
```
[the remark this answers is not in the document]
hatesalp: It should stay this way.

src/lib/components/AlertModal/AlertModal.test.tsx
```diff
@@ -0,0 +1,200 @@
...
+  });
+
+  it("should render the modal when open is true", () => {
+    const { rerender, queryByTestId, getBackdrop } = renderExt(<AlertModal title="Alert Modal Title" open={false} />);
+    expect(queryByTestId("alertModalBackdrop")).not.toBeInTheDocument();
+    rerender(<AlertModal title="Alert Modal Title" open />);
+    expect(getBackdrop()).toBeInTheDocument();
+  });
+
+  it("should render the title given in the title prop", () => {
```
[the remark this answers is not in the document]
hatesalp: It will be updated in the next commit.

src/lib/components/AlertModal/AlertModal.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const { visible, show, hide } = useToggle(open);
+
+  return (
+    <>
+      <Button label="Open Modal" onClick={show} />
+      <AlertModal
+        open={visible}
+        onClose={hide}
+        icon={<MotifIcon name="info" size="xxl" />}
+        actionButton={{ text: "Confirm", onClick: () => alert("Action clicked!") }}
```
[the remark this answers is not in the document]
hatesalp: It will be updated in the next commit.

src/lib/components/AlertModal/AlertModal.tsx
```diff
@@ -0,0 +1,95 @@
...
+    enableDivider = false,
+    bordered,
+    elevated,
+    variant = "primary",
+    ref,
+    style,
+    className,
+  } = usePropsWithThemeDefaults("AlertModal", props);
+
+  const domReady = useDomReady();
```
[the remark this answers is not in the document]
hatesalp: This part may change. You can ignore it for now.

src/lib/components/AlertModal/AlertModal.test.tsx
```diff
@@ -0,0 +1,200 @@
...
+    getBackdrop,
+    getModalActions,
+    getModalContent,
+  };
+};
+
+describe("AlertModal", () => {
+  it("should be rendered with only required props and should have default prop values stated here", () => {
+    const { container, getBackdrop, getModalActions } = renderExt(
+      <AlertModal title="Alert Modal Title" open icon={<MotifIcon name="home" />} actionButton={{ text: "Confirm", onClick: jest.fn() }} />,
```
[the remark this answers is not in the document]
hatesalp: The intent of this test is to assert that the default variant="primary" propagates correctly to child elements. Since icon and actionButton are the only props that visually reflect the variant, they're needed here as vehicles not as feature coverage. The variant test below iterates all variants but doesn't assert the default (no variant prop passed), so this test covers a distinct case: default variant without explicit prop.

src/lib/components/Alert/Alert.tsx
```diff
@@ -19,12 +19,12 @@ const Alert = (props: PropsWithRefAndChildren<AlertProps, HTMLDivElement>) => {
     style,
   } = usePropsWithThemeDefaults("Alert", props);
 
-  const { visible, hide, toggleState } = useToggle(true, 300);
+  const { visible, hide, toggleState } = useToggle(true, { showTime: 300, hideTime: 300 });
```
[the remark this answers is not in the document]
hatesalp: updated

src/lib/components/AlertModal/AlertModal.tsx
```diff
@@ -0,0 +1,90 @@
...
+  attachedRef.current = attached;
+
+  const prevToggleState = useRef(toggleState);
+  useEffect(() => {
+    if (prevToggleState.current === "hiding" && !toggleState) onClose?.();
+    prevToggleState.current = toggleState;
+  }, [toggleState, onClose]);
+
+  const modalRef = useOutsideClick<HTMLDivElement>(() => {
+    onClose && hide();
```
aktasmehmet: it would be better to connect hiding to an external prop like "closable" instead of depending on onClose. This prop should be put into defaultable props

src/lib/components/AlertModal/AlertModal.module.scss
```diff
@@ -0,0 +1,124 @@
...
+  background-color: var(--base-colors-black-black-overlay50);
+}
+
+.elevated .alertModalContainer {
+  @include shadow(6);
+}
+
+.bordered .alertModalContainer {
+  border: var(--base-sizing-1x) solid var(--theme-color-border-light-default);
+}
```
aktasmehmet: You can move these 2 to group inside .alertModalContainer section like:

```
.alertModalContainer {                                                                                                                                                                                                                                                                                                    
    display: flex;                                                                                                                                                                                                                                                                                                          
    flex-direction: column;                                                                                                                                                                                                                                                                                                 
    background: var(--theme-color-surface-base-bg-light);                                                                                                                                                                                                                                                                   
    color: var(--theme-color-text-base-title);                                                                                                                                                                                                                                                                              
    padding: var(--base-sizing-8x);                                                                                                                                                                                                                                                                                         
    border-radius: var(--theme-sizing-radius-default);                                                                                                                                                                                                                                                                      
    @include pop-default;                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                            
    .elevated & {                                                                                                                                                                                                                                                                                                           
      @include shadow(6);                                                                                                                                                                                                                                                                                                   
    }                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                            
    .bordered & {                                                                                                                                                                                                                                                                                                           
      border: var(--base-sizing-1x) solid var(--theme-color-border-light-default);                                                                                                                                                                                                                                          
    }                                                                                                                                                                                                                                                                                                                       
  } 
```

src/lib/components/AlertModal/AlertModal.tsx
```diff
@@ -0,0 +1,90 @@
...
+    icon,
+    size = "md",
+    open,
+    onClose,
+    actionButton,
+    alternateButton,
+    buttonsPosition = "center",
+    contentPosition = "center",
+    removeBackdrop,
+    enableDivider = false,
```
aktasmehmet: is default value here necessary?

src/lib/components/AlertModal/AlertModal.test.tsx
```diff
@@ -0,0 +1,197 @@
...
+  it("should be rendered with only required props and should have default prop values stated here", () => {
+    const { container, getBackdrop, getModalActions } = renderExt(
+      <AlertModal title={TITLE} open icon={<MotifIcon name="home" />} actionButton={{ text: "Confirm", onClick: jest.fn() }} />,
+    );
+    expect(container).toMatchSnapshot();
+
+    // size: md
+    expect(getBackdrop()).toHaveClass("md");
+    // removeBackdrop: false
+    expect(getBackdrop()).toHaveClass("backdrop");
```
aktasmehmet: this is a default behaviour test, not prop. I can also test lots of css classes here.
So you have to remove this one.

src/lib/components/AlertModal/AlertModal.test.tsx
```diff
@@ -0,0 +1,197 @@
...
+    expect(handleConfirm).toHaveBeenCalledTimes(1);
+    expect(handleCancel).toHaveBeenCalledTimes(1);
+  });
+
+  it("should not render the actions area when neither action button nor alternate button is provided", () => {
+    renderExt(<AlertModal title={TITLE} open />);
+    expect(screen.queryByRole("button")).not.toBeInTheDocument();
+  });
+
+  it("should not render divider by default when buttons are provided", () => {
```
aktasmehmet: this should be moved into the first test

src/lib/components/AlertModal/AlertModal.test.tsx
```diff
@@ -0,0 +1,197 @@
...
+  });
+
+  it("should render action button and call action handler when clicked", () => {
+    const handleAction = jest.fn();
+    renderExt(<AlertModal title={TITLE} open actionButton={{ text: "Confirm", onClick: handleAction }} />);
+    fireEvent.click(screen.getByText("Confirm"));
+    expect(handleAction).toHaveBeenCalledTimes(1);
+  });
+
+  it("should call each handler when action button and alternate button are clicked", () => {
```
aktasmehmet: functioning at the same time is not a necessary test scenario. You may remove this.

src/lib/components/AlertModal/AlertModal.test.tsx
```diff
@@ -0,0 +1,197 @@
...
+    const positions: AlertModalButtonPosition[] = ["left", "center", "right", "spaceBetween", "stretch", "fullWidth"];
+    positions.forEach(buttonsPosition => {
+      const { getBackdrop } = renderExt(
+        <AlertModal title={TITLE} open buttonsPosition={buttonsPosition} actionButton={{ text: "Confirm", onClick: jest.fn() }} />,
+      );
+      expect(getBackdrop()).toHaveClass(`actions_${buttonsPosition}`);
+      cleanup();
+    });
+  });
+  it("should render with the given variant class on icon and action button", () => {
```
aktasmehmet: This may include more in the future. I dont know, background or something. So it would be better to rename as:


```suggestion
  it("should render with the given variant", () => {
```

src/lib/components/AlertModal/AlertModal.test.tsx
```diff
@@ -0,0 +1,197 @@
...
+  it("should apply the given contentPosition class", () => {
+    const contentPosition: AlertModalContentPosition[] = ["left", "center", "right"];
+    contentPosition.forEach(contentPosition => {
+      const { getBackdrop } = renderExt(<AlertModal title={TITLE} open contentPosition={contentPosition} />);
+      expect(getBackdrop()).toHaveClass(`content_${contentPosition}`);
+      cleanup();
+    });
+  });
+
+  it("should apply the given buttonsPosition class to the actions area", () => {
```
aktasmehmet:
```suggestion
  it("should position the buttons as given in buttonsPosition prop", () => {
```

src/lib/components/AlertModal/AlertModal.test.tsx
```diff
@@ -0,0 +1,197 @@
...
+  });
+
+  it("should render alternate button and call handler when clicked", () => {
+    const handleCancel = jest.fn();
+    renderExt(<AlertModal title={TITLE} open alternateButton={{ text: "Cancel", onClick: handleCancel }} />);
+    fireEvent.click(screen.getByText("Cancel"));
+    expect(handleCancel).toHaveBeenCalledTimes(1);
+  });
+
+  it("should apply the given contentPosition class", () => {
```
aktasmehmet:
```suggestion
  it("should position the content given in contentPosition prop", () => {
```

src/lib/components/AlertModal/AlertModal.test.tsx
aktasmehmet: There are still missing tests:

- ``open`` prop
- icon and icon position for alternate and action buttons

src/lib/components/AlertModal/AlertModal.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+    contentPosition: { table: { defaultValue: { summary: "center" } } },
+    buttonsPosition: { table: { defaultValue: { summary: "center" } } },
+    size: { table: { defaultValue: { summary: "md" } } },
+    variant: { table: { defaultValue: { summary: "primary" } } },
+  },
+  args: {
+    title: "Alert Modal Title",
+    text: "Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industrys standard dummy text ",
+    contentPosition: "center",
+    buttonsPosition: "center",
```
aktasmehmet: no need for these 2. They are already default values

src/lib/components/AlertModal/AlertModal.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const { open, ...otherProps } = props;
+  const { visible, show, hide } = useToggle(open);
+
+  return (
+    <>
+      <Button label="Open Modal" onClick={show} />
+      <AlertModal
+        open={visible}
+        onClose={hide}
+        icon={<MotifIcon name="info" size="xxl" />}
```
aktasmehmet: icon should be provided from icon args and remove here. you can only use string.

src/lib/components/AlertModal/AlertModal.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+        alternateButton={{ text: "Close", onClick: hide }}
```
aktasmehmet: alternateButton should be provided from the args and removed from here

aktasmehmet (changes requested):

src/lib/components/AlertModal/AlertModal.test.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+    jest.useRealTimers();
+  });
+
+  it("should not close the modal when clicked outside of it when closable prop is false", () => {
+    const handleClose = jest.fn();
+    const { getBackdrop } = renderExt(<AlertModal title={TITLE} open closable={false} onClose={handleClose} />);
+    fireEvent.click(getBackdrop());
+    expect(handleClose).not.toHaveBeenCalled();
+  });
```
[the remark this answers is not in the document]
hatesalp: closable && hide() treats both undefined and false identically neither calls hide(). The existing closable={false} test already covers this exact code path. Modal follows the same pattern and only tests explicit true/false cases; aligning with that convention here as well

src/lib/hooks/useToggle.ts
```diff
@@ -11,33 +16,48 @@ type UseToggleReturn = {
...
 
+  const pendingRafRef = useRef<number[]>([]);
+  const pendingTimeoutRef = useRef<ReturnType<typeof setTimeout> | undefined>(undefined);
+
+  // To prevention quick show/hide that may change already settled frame
+  const clearPending = useCallback(() => {
+    pendingRafRef.current.forEach(id => cancelAnimationFrame(id));
+    pendingRafRef.current = [];
+    clearTimeout(pendingTimeoutRef.current);
+  }, []);
```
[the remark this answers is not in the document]
hatesalp: updated

src/lib/hooks/useToggle.ts
```diff
@@ -11,33 +16,34 @@ type UseToggleReturn = {
   toggle: (forceShow?: boolean) => void;
 };
 
-const useToggle = (initialVisible = false, toggleTime?: number): UseToggleReturn => {
+const useToggle = (options: Options = {}): UseToggleReturn => {
```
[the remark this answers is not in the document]
hatesalp: We'll handle it this way for now.

atakanbayrak (dismissed):

aktasmehmet (approved):

atakanbayrak (approved):

PR #144, opened 2026-08-26
refactor(UploadInput): add conditions for showing the download button

hatesalp:
## Description

- Download button is now shown only when onDownloadClick is provided and the file is in SUCCESS or DELETE_FAIL state
- Removed the unused hasSuffix_ padding mechanism and focusable class from LabelArea
- Removed unused class props from LabelSuffix button rendering

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [x] 🧹 Code Refactoring

## Screenshots / Preview

{4B7BE4A2-BCA9-4AD5-9BBC-E7569F5E57E6}
## Checklist


src/lib/components/Upload/UploadInput/UploadInput.module.scss
```diff
@@ -159,11 +159,6 @@ $sizes: (("xs", "4x", "4x", "10x"), ("sm", "6x", "6x", "12x"), ("md", "8x", "8x"
     .label {
       @include font("body", $size);
       padding: var(--base-sizing-#{$paddingV}) var(--base-sizing-#{$paddingH});
```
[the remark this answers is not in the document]
aktasmehmet: +1 on that. Can you check please?

src/lib/components/Upload/UploadInput/UploadInput.module.scss
aktasmehmet, line 146: Claude's findings look valid. Can you check please?

hatesalp: Sure, I'm on it. I'll update it shortly.

src/lib/components/Upload/UploadInput/UploadInput.module.scss
```diff
@@ -132,11 +132,9 @@ $sizes: (("xs", "4x", "4x", "10x"), ("sm", "6x", "6x", "12x"), ("md", "8x", "8x"
   }
 
   .labelSuffixWrapper {
```
aktasmehmet, line 124: In general, the whole cleanup is fine, you've also found some unused classes, great.

Here are some bugs I've found:

1 - There is a hover state in the label area which causes unnecessary border render there:

https://github.com/user-attachments/assets/498c21d3-76ea-44e6-954a-dabd6c648e77

The thing is, I've checked, that hover on line #130 is not necessary at all. I guess you can remove that.

2 - Since you've given up absolute positioning (which is fine) and use padding, instead of right... This cause an unnecessary empty space when there is no suffixes at all. In fact, suffix container should not be rendered at all (no item in it, check ss). You should fix here

Image
Image

3 - Please check the gap between the label and the buttons in figma. Here in the code, it is not the same and they are next to each other without gap at all

Image
Image

aktasmehmet (changes requested):

src/lib/components/Upload/UploadInput/UploadInput.module.scss
hatesalp: Thanks! I’ll push a commit now, and you can check it afterward.

src/lib/components/Upload/UploadInput/UploadInput.module.scss
ZehranurC: No longer needed here (no absolute children),  i think it can be removed.

hatesalp: I think this is outside the scope of the current task, so I’d prefer to leave it as is for now.

aktasmehmet: I don't think so. It is somewhat related. Besides, even if it is not, this kind of very very minor leftovers should be fixed immediately. And in fact, we should thank @ZehranurC for finding it :)

src/lib/components/Upload/UploadInput/UploadInput.test.tsx
```diff
@@ -546,4 +544,9 @@ describe("UploadInput", () => {
     await waitFor(() => expect(getDownloadButton()).not.toBeInTheDocument());
     xhrSpy.mockRestore();
   });
+
+  it("should not render the suffix area when there is no file and no error or success prop", () => {
```
aktasmehmet: "suffix area" is not an understandable naming. Please explictly refer what should not render here.

src/lib/components/Upload/UploadInput/UploadInput.module.scss
aktasmehmet: There are still some issues (even if it looks fine as a result):

- Margin should be preferred instead of padding. 
- More importantly, check the visible gap between the input text area and the icons/buttons. Previously, when they were absolutely positioned, we were using the padding of the input itself. But now, if you keep that, that will cause an additional gap between the text and the icons. Because now the input is not rightmost aligned. (When fixing this, please pay extra attention of the scenario that there is no icons/buttons at all.)

Image
Image
Image

Image

aktasmehmet (changes requested):

src/lib/components/Upload/UploadInput/UploadInput.module.scss
hatesalp: You're right. It will be updated in the next commit.

src/lib/components/Upload/UploadInput/UploadInput.module.scss
hatesalp: This is how it looks after the latest update.

{3C760428-4EB7-4FE4-A722-59C294D44E2A}

{71EE7D87-6548-466D-98C1-201CFEA4CEFB}

{59E2A336-7FDF-4024-AB28-3DF43586FD0D}

src/lib/components/Upload/UploadInput/UploadInput.module.scss
```diff
@@ -126,17 +125,12 @@ $sizes: (("xs", "4x", "4x", "10x"), ("sm", "6x", "6x", "12x"), ("md", "8x", "8x"
...
-    position: absolute;
     display: flex;
     background: none;
-    top: 50%;
-    transform: translateY(-50%);
+    align-self: center;
     border-radius: var(--theme-sizing-radius-default);
     border: solid var(--base-sizing-1x) transparent;
     @include transition(color);
```
aktasmehmet: since these are not absolute now and aligned to the edge, border transparent, radius, etc are all redundant. You can also remove the transition here, buttons can handle that themselves.

src/lib/components/Upload/UploadInput/UploadInput.module.scss
```diff
@@ -158,15 +152,15 @@ $sizes: (("xs", "4x", "4x", "10x"), ("sm", "6x", "6x", "12x"), ("md", "8x", "8x"
...
+      padding: var(--base-sizing-#{$paddingV}) 0 var(--base-sizing-#{$paddingV}) var(--base-sizing-#{$paddingH});
+
+      &:last-child {
+        padding-right: var(--base-sizing-#{$paddingH});
       }
     }
     .labelSuffixWrapper {
-      right: var(--base-sizing-#{$paddingH});
+      margin-right: var(--base-sizing-#{$paddingH});
+      margin-left: var(--base-sizing-#{$paddingV});
```
aktasmehmet: instead using margin-left here, you may handle this one level up, in `labelArea` by giving a gap there?

aktasmehmet (approved):

ZehranurC (dismissed):

aktasmehmet (dismissed):

src/lib/components/Upload/UploadInput/UploadInput.module.scss
```diff
@@ -1,7 +1,7 @@
 @use "@styles" as *;
 
-// size, gap, paddingV, paddingH
-$sizes: (("xs", "4x", "4x", "10x"), ("sm", "6x", "6x", "12x"), ("md", "8x", "8x", "16x"), ("lg", "8x", "10x", "20x"));
+// size, gap&paddingV, paddingH
+$sizes: (("xs", "4x", "6x"), ("sm", "6x", "12x"), ("md", "8x", "16x"), ("lg", "10x", "20x"));
```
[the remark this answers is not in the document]
aktasmehmet: this change is by design, it was a bug before

ZehranurC (approved):

PR #146, opened 2026-08-27
refactor: Toast and UploadList refactoring fixes

atakanbayrak:
## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [x] 🧹 Code Refactoring

## Checklist

### DIFF
--- a/src/lib/components/Upload/UploadFileList/components/Row/FileButton.tsx
+++ b/src/lib/components/Upload/UploadFileList/components/Row/FileButton.tsx
@@ -4,6 +4,7 @@ import { MotifIconButton } from "@/components/Motif/Icon";
 import { memo, useContext } from "react";
 import { FileType } from "@/components/Upload/types";
 import { UploadContext } from "@/components/Upload/UploadProvider";
+import IconButton from "@/components/IconButton";
 
 type Props = {
   file: FileType;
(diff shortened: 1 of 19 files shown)


src/lib/components/Upload/UploadFileList/components/Row/FileButton.tsx
```diff
@@ -18,7 +19,7 @@ export const FileButton = memo(({ file, readOnly, disabled }: Props) => {
   return (
     <>
       {file.action && (
-        <MotifIconButton
+        <IconButton
```
aktasmehmet, line 22: Why this change?

aktasmehmet (changes requested):
There are still missing refactors:

1 - In UploadDragger story file, value prop type definition does not have `action`
2 - test came with action in UploadList is missing in UploadDragger

src/lib/components/Upload/UploadFileList/components/Row/FileButton.tsx
atakanbayrak: IconGlobalType is usable for IconButton previously MotifIconButton was usable because I implemented as string, same example appears in CardHeader as

```code
{action && <IconButton name={action.icon} onClick={action.onClick} size="xxl" variant={iconButtonVariant} />}
```

aktasmehmet: Because the CardHeader has a user defined icon (in action), In upload, those icons are predefined, non-changable. So, as you know, those type of icons are the main reason that we created MotifIconButton.

In fact, this was not something requested at all. You need to revert that

aktasmehmet (approved):

PR #150, opened 2026-09-03
test: add className, style and ref props test to all components

ZehranurC:
## Description

- Added a runStandardPropsTest helper and applied it across all component tests to verify className/style/ref reach the root element.
- Form: test written manually instead of using the helper, since its ref is a FormRefType imperative handle, not a DOM ref to the root <form>.
- Slider: fixed its exported prop type, which was missing className/style/ref (it already worked at runtime, only the type was incomplete).
- UploadInput: className/style/ref are now correctly forwarded to the root element.

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [x] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [x] 🧹 Code Refactoring

## Screenshots / Preview

image
## Checklist

## How to Test

-- npm  run test:ci


src/utils/testUtils.tsx
```diff
@@ -29,6 +31,35 @@ export const runIconPropTest = (renderIcon: ComponentWithIconRender, className?:
   testStringIcon();
   testReactElementIcon();
 };
+
+// Root element locator; override for portals (e.g. Modal, Popover, Toast). Defaults to container.firstElementChild.
+export const runStandardPropsTest = <E extends Element = HTMLElement>(
+  renderComponent: (props: { className?: string; style?: CSSProperties; ref?: Ref<E> }) => RenderResult,
```
atakanbayrak: We may consider to make props not optional if we check them for each component.

aktasmehmet: Why? className, style and ref are optional in all of the components? props itself is not optional here. Am I missing something?

src/utils/testUtils.tsx
```diff
@@ -29,6 +31,34 @@ export const runIconPropTest = (renderIcon: ComponentWithIconRender, className?:
...
+  const testRef = () => {
+    const ref = createRef<E>();
+    const result = renderComponent({ ref });
+    expect(ref.current).toBe(getRoot(result));
+    result.unmount();
+  };
+
+  testClassName();
+  testStyle();
+  testRef();
```
aktasmehmet: Creating 3 different functions is not necessary here, since we are not using them seperately. So, you may delete the functions and keep whats inside? In fact, you can also merge custom class, style and ref into 1 render?

src/lib/components/Textarea/Textarea.test.tsx
```diff
@@ -2,12 +2,17 @@ import "@testing-library/jest-dom";
...
 
 describe("Textarea", () => {
   it("should render with only required props", () => {
     expect(render(<Textarea />).container).toMatchSnapshot();
   });
 
+  it("should render with the given className, style and ref on the root element", () => {
+    runStandardPropsTest<HTMLDivElement>(props => render(<Textarea {...props} />));
+  });
+
```
aktasmehmet: **!!! It is for all of the tests !!!**

We alreay have one, standard test to put in all of the test files: "required props test". So, why dont you rename them and put and run this runStandardPropsTest inside them.

This way it will reduce the number of tests not to forget in the future :)

You may rename them as 

``should be rendered with only required props, should have default prop values stated here, and have standard props; className, style and ref props working as expected```

or 

``should render with only required props, and have standard props; className, style and ref props working as expected``

aktasmehmet (changes requested):

src/lib/components/Textarea/Textarea.test.tsx
ZehranurC: - Merged the separate runStandardPropsTest test into each component's existing "required props" test — one test per file now, not two.
- Renamed them using your suggested titles: "...should have default prop values stated here, and have standard props; className, style and ref props working as expected" for components whose test also checks default values, and the shorter variant for the rest — applied consistently across all 60 test files.
- Two exceptions, documented with a comment in the test: Form (ref is an imperative handle, not a DOM node) and originally Tooltip (portal mounts only on hover) — Tooltip was later fixed too and now uses runStandardPropsTest like everyone else.

src/lib/components/IconButton/IconButton.test.tsx
```diff
@@ -6,10 +6,12 @@ import { MOTIF_ICONS_DEFAULT_CLASS } from "../../constants";
 import { Size7 } from "../../types";
 import Icon from "@/components/Icon";
 import MotifProvider from "../../motif/context/MotifProvider";
+import { runStandardPropsTest } from "../../../utils/testUtils";
 
 describe("IconButton", () => {
-  it("should be rendered with only required props", () => {
+  it("should render with only required props, and have standard props; className, style and ref props working as expected", () => {
     expect(render(<IconButton name="account_circle" />).container).toMatchSnapshot();
+    runStandardPropsTest<HTMLButtonElement>(props => render(<IconButton name="account_circle" {...props} />));
```
aktasmehmet: There are duplication everywhere. For example, here: ``<IconButton name="account_circle" />``. If there is a change in the future, we need to change 2 places. Instead, you can follow a solujtion like this:

```
const renderComponent = (props = {}) => render(<IconButton name="account_circle" {...props} />);
expect(renderComponent().container).toMatchSnapshot();
runStandardPropsTest<HTMLButtonElement>(renderComponent);
```

- In fact, you can even create anohter function but them inside.
- Even further: The test names are identical. You can even extract the whole `it()` block out of these files and create a function to process them all including the "it".

Like: 

```
const runSnapshotAndStandardPropsTests = (/*RENDER HERE*/) => {
  it("should render with only required props, and have standard props; className, style and ref props working as expected", () => {
    // LOGIC HERE
  };
};
```

aktasmehmet (changes requested):

src/lib/components/Tooltip/Tooltip.test.tsx
```diff
@@ -38,6 +38,26 @@ describe("Tooltip", () => {
...
+    render(
+      <Tooltip text="Description" className="custom-class" style={{ marginTop: "13px" }}>
+        <Button label="Test Button" />
+      </Tooltip>,
+    );
+
+    await act(() => user.hover(screen.getByText("Test Button")));
+    const tooltip = screen.getByTestId("tooltipItem");
+
+    expect(tooltip).toHaveClass("custom-class");
```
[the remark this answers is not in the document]
ZehranurC: This was updated in the new commit.

src/utils/testUtils.tsx
ZehranurC: This was updated in the new commit.

src/lib/components/IconButton/IconButton.test.tsx
ZehranurC: Done, extracted into a shared runSnapshotDefaultsAndStandardPropsTest helper that wraps the whole it() block, now used across all test files.

src/lib/components/DateTimePicker/DateTimePicker.test.tsx
```diff
@@ -44,6 +44,8 @@ describe("DateTimePicker", () => {
 
   runTimePickerCommonTests();
 
+  runSnapshotDefaultsAndStandardPropsTest((props: StandardProps<HTMLDivElement>) => render(<DateTimePicker {...props} />));
```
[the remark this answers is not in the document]
hatesalp: I encountered the same issue while updating the snapshots. The DateTimePicker snapshot changes according to the current date because it is rendered without a fixed value. The system time should be fixed in the test so that it always produces the same snapshot

src/lib/components/Pagination/Pagination.test.tsx
```diff
@@ -9,6 +10,10 @@ describe("Pagination", () => {
     mockOnChange.mockClear();
   });
 
+  runSnapshotDefaultsAndStandardPropsTest((props: StandardProps<HTMLDivElement>) =>
```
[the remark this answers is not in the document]
ZehranurC: Redundant tests were removed in the last commit.

src/lib/components/DateTimePicker/DateTimePicker.test.tsx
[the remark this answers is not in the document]
ZehranurC: I'm not sure we should fix this here. Maybe we can address it later.

src/lib/components/Carousel/Carousel.test.tsx
```diff
@@ -2,10 +2,11 @@ import "@testing-library/jest-dom";
 import { render, screen, fireEvent, act } from "@testing-library/react";
 import { userEvent } from "@testing-library/user-event";
 import Carousel from "./Carousel";
-import { IndicatorShape, Theme } from "./types";
+import { CarouselProps, IndicatorShape, Theme } from "./types";
 import { ReactNode } from "react";
+import { runSnapshotDefaultsAndStandardPropsTest, StandardProps } from "../../../utils/testUtils";
 
-const renderExt = (props = {}, component?: ReactNode) => {
+const renderExt = (props: CarouselProps & StandardProps<HTMLDivElement> = {}, component?: ReactNode) => {
```
aktasmehmet: This use is redundant. We already have a type that covers standard props alongside with other props: PropsWithRef.
So you can replace here as:

```suggestion
const renderExt = (props: PropsWithRef<CarouselProps, HTMLDivElement> = {}, component?: ReactNode) => {
```

src/lib/components/Stepper/Stepper.test.tsx
```diff
@@ -9,7 +9,7 @@ const NEXT = t("g.next");
 const PREV = t("g.previous");
 const FINISH = t("g.finish");
 
-const renderExt = (props: StepperProps = {}, itemProps?: StepperItemProps[]) => {
+const renderExt = (props: StepperProps & StandardProps<HTMLDivElement> = {}, itemProps?: StepperItemProps[]) => {
```
aktasmehmet: This use is redundant. We already have a type that covers standard props alongside with other props: PropsWithRef.
So you can replace here as:

```suggestion
const renderExt = (props: PropsWithRef<StepperProps, HTMLDivElement> = {}, component?: ReactNode) => {
```

src/utils/testUtils.tsx
```diff
@@ -29,6 +31,46 @@ export const runIconPropTest = (renderIcon: ComponentWithIconRender, className?:
...
+  className?: string;
+  style?: CSSProperties;
+  ref?: Ref<E>;
+};
+
+export type StandardPropsRender<E extends Element = HTMLElement, R extends RenderResult = RenderResult> = (props: StandardProps<E>) => R;
+
+/**
+ * @param renderComponent renders the component, spreading the given `props` onto it
+ * @param options.assertDefaults extra checks on the plain render; when given, the test title also states that default prop values are covered
```
aktasmehmet, line 37: You should add ``@param options`` on top of this line. It gives a lint warning.

src/utils/testUtils.tsx
```diff
@@ -29,6 +31,46 @@ export const runIconPropTest = (renderIcon: ComponentWithIconRender, className?:
...
+ * @param options.getRoot locates the root element for the contract asserts (default `container.firstElementChild`; override for portals)
+ */
+export const runSnapshotDefaultsAndStandardPropsTest = <E extends Element = HTMLElement, R extends RenderResult = RenderResult>(
+  renderComponent: StandardPropsRender<E, R>,
```
aktasmehmet: I tested that we can remove StandardProps and StandardPropsRender types here and just create another helper type in the main src/lib/types.ts file (instead of this file) called StandardPropsWithRef and use it here as:


```suggestion
  renderComponent: (props: StandardPropsWithRef<E>) => R,
```

I'll also drop the suggestion to types.ts file here. There are other little refactors as well but the final version of the lines with `PropsWithRef`, `PropsWithRefAndChildren`, `StandardPropsWithChildren` should be revised as:

```
export type PropsWithRef<P = unknown, R = unknown> = P & StandardPropsWithRef<R>;
export type PropsWithRefAndChildren<P = unknown, R = unknown> = PropsWithChildren<PropsWithRef<P, R>>;
export type StandardPropsWithRef<R = unknown> = StandardProps & RefType<R>;
export type StandardPropsWithChildren<P = unknown> = PropsWithChildren<P> & StandardProps;
```

aktasmehmet (changes requested):
There is a little bit more refactoring, then it is done. Good work 💪

aktasmehmet (approved):

PR #151, opened 2026-09-04
chore(deps): TypeScript version 10 upgrade

atakanbayrak:
## Description

New rules that came in automatically with @eslint/js's recommended config on ESLint 10:

  - no-unassigned-vars — flags let/var declarations that are never assigned a value anywhere.
  - no-useless-assignment — flags an assignment whose value is never read before being overwritten or going out of scope (a dead assignment).
  - preserve-caught-error — flags re-throwing a new error inside a catch block without preserving the original error (e.g. via cause).

None of these three flagged anything in the existing codebase — npm run lint passes clean.

The @typescript-eslint/* bump (8.67.0 → 8.69.0) is a minor/patch release, so recommendedTypeChecked's rule set is unchanged — no new TypeScript rules came with it, it was just bumped alongside ESLint for v10 compatibility.

- no-shadow (or @typescript-eslint/no-shadow for a TS project) — catches variable shadowing, an easy mistake in large component files.
- array-callback-return — catches a missing return in map/filter/reduce callbacks.
- max-params — caps the number of function parameters, useful for readability in prop-heavy components.

Tried enabling all three, but didn't find them necessary for this repo at this time — leaving them out of scope for this ESLint 10 upgrade PR.

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Checklist


atakanbayrak:
"eslint-plugin-react": "7.37.5" does not support TS10 yet. 

https://github.com/jsx-eslint/eslint-plugin-react/issues/3977

^^^

Will be checked in advance.

PR #152, opened 2026-09-04
docs(Storybook): hide form component horizontal scrollbar

hatesalp:
## Description

`Storybook Form Component (horizontal):` Changed overflow: scroll to overflowX: scroll to prevent the unnecessary vertical scrollbar from appearing on Windows. Only horizontal scrolling is enabled.
## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [x] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Screenshots / Preview

BEFORE: 
image

AFTER: 
image

## Checklist

### DIFF
--- a/src/lib/components/Form/Form.stories.tsx
+++ b/src/lib/components/Form/Form.stories.tsx
@@ -35,7 +35,7 @@ const meta: Meta<typeof Form> = {
     (Story, context) => (
       <div
         style={{
-          ...(context.args.formOrientation === "horizontal" ? { width: "100%", overflow: "scroll" } : { width: 500, overflow: "initial" }),
+          ...(context.args.formOrientation === "horizontal" ? { width: "100%", overflowX: "scroll" } : { width: 500, overflow: "initial" }),
           padding: 50,
           margin: "0 auto",
         }}


src/lib/components/Form/Form.stories.tsx
```diff
@@ -35,7 +35,7 @@ const meta: Meta<typeof Form> = {
     (Story, context) => (
       <div
         style={{
-          ...(context.args.formOrientation === "horizontal" ? { width: "100%", overflow: "scroll" } : { width: 500, overflow: "initial" }),
+          ...(context.args.formOrientation === "horizontal" ? { width: "100%", overflowX: "scroll" } : { width: 500, overflow: "initial" }),
```
atakanbayrak, line 38: I am not sure this decision was maden by design team or our team, but I didn't understand the benefits of this situation. In the end, we didn't manage to solve scroll bar appearing issue and we only disabled for Y axis. If there is no solution at all, we could also revert this change in my opinion.

aktasmehmet: Actually I've talked with @hatesalp about it and this is a small visible improvement for just this story in particular.

After thinking more over this, I also agree @atakanbayrak . This is not needed at all. In the end, there it looks like there is no solution to fix this windows in particular. 

So, it would be better to revert this, delete the PR and resolve the task 🤷‍♂️

PR #154, opened 2026-09-04
fix(Storybook): mobile and responsive view issues 

aktasmehmet:
## Description

This PR aims to solve recently detected mobile/responsive view issues in storybook. Broken views are added to the Screenshot section below.

## Type of Change

- [ ] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [x] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [x] 🧹 Code Refactoring

## Screenshots / Preview

**ERRORS**

New Project

## Checklist

## How to Test

Mostly based on storybook. Yu should use mobile view in browser or more preferably a real mobile device.

### DIFF
--- a/.storybook/manager-head.html
+++ b/.storybook/manager-head.html
@@ -90,6 +90,12 @@
       border-right: 1px solid var(--mtfdocs-border);
     }
   }
+  /* ---------------------------
+        Sidebar - Mobile Menu
+    --------------------------- */
+  #storybook-mobile-menu div:has(> #storybook-explorer-menu) {
+    padding-bottom: 64px;
+  }
 
   /* ---------------------------
         Sidebar - Nav Items
--- a/.storybook/manager-head.html
+++ b/.storybook/manager-head.html
@@ -93,8 +93,17 @@
   /* ---------------------------
         Sidebar - Mobile Menu
     --------------------------- */
+  /* iOS Safari bottom address bar, menu cut off fix */
+  #storybook-mobile-menu {
+    height: 80vh !important;
+  }
+  @supports (height: 100dvh) {
+    #storybook-mobile-menu {
+      height: 80dvh !important;
+    }
+  }
   #storybook-mobile-menu div:has(> #storybook-explorer-menu) {
-    padding-bottom: 64px;
+    padding-bottom: calc(16px + env(safe-area-inset-bottom) + (100lvh - 100svh));
   }
 
   /* ---------------------------
@@ -226,6 +235,10 @@
 
       &:hover {
         background: rgba(var(--mtfdocs-purple-rgb), 0.14);
+      }
+      /* The mobile bottom bar's nav-menu crop fix */
+      p {
+        line-height: 1.4;
       }
     }
     button[aria-label*="Global theme for components"] {
--- a/src/i18n/locales/en.json
+++ b/src/i18n/locales/en.json
@@ -23,7 +23,7 @@
   },
   "table": {
     "totalSelectedRecords": "{{selected}} of {{total}} records selected.",
-    "totalRecords": "Total {{total}} records available."
+    "totalRecords": "Total {{total}} records"
   },
   "validation": {
     "required": "Please fill in this field",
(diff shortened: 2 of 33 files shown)


src/lib/components/Select/Select.module.scss
```diff
@@ -68,9 +68,7 @@ $sizes: (
 
   .inputText {
     all: unset;
-    width: 0;
-    min-width: 0;
-    flex-grow: 1;
+    width: 100%;
     text-align: left;
     white-space: nowrap;
```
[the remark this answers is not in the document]
atakanbayrak: Better to check twice.

src/i18n/locales/en.json
```diff
@@ -23,7 +23,7 @@
   },
   "table": {
     "totalSelectedRecords": "{{selected}} of {{total}} records selected.",
-    "totalRecords": "Total {{total}} records available."
+    "totalRecords": "Total {{total}} records"
```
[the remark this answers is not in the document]
atakanbayrak: Better to check it again.

src/lib/components/Table/Table.stories.tsx
ZehranurC: This doesn't match the actual render: the render uses position: "center" + hideTotalRecords, but the displayed snippet still shows position: "right" and no hideTotalRecords. "Show code" ends up being misleading. Can you sync it up?

src/i18n/locales/en.json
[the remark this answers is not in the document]
ZehranurC: +1 --> It doesn't seem consistent with tr.json.

.storybook/manager-head.html
```diff
@@ -90,6 +90,21 @@
       border-right: 1px solid var(--mtfdocs-border);
     }
   }
+  /* ---------------------------
```
hatesalp, line 93: I wasn't sure where to leave these comments, so I'm adding them at the beginning.

Nice work.  However, I noticed a few points:

1. On the **Design Themes** page, some components overlap, and a horizontal scrollbar appears when they don't fit. If we stack the components vertically instead of forcing them to stay side by side, we can avoid the horizontal scrollbar and also fix the overlapping issue.

2. The **InputDateTime** picker gets clipped when it opens.
3. The `width: 900px` applied to the `Grid Col` seems to create unnecessary empty space.

Image

Image

Image

Image

Image

Image

src/lib/components/Select/Select.module.scss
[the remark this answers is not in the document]
aktasmehmet: I've checked and found the issue. It is fixed and the fix will be applied in the next commit. Resolving.

src/i18n/locales/en.json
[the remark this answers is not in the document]
aktasmehmet: Correct, just fixed. Coming in next commit.

.storybook/manager-head.html
aktasmehmet: > 2. The InputDateTime picker gets clipped when it opens.

I am aware of that. Fixe the clipping issue with a workaround by enlarging the container. This is a temporary fix to see the component working. This issue will be completely resolved when those overlay component will switch to portaling to body.

aktasmehmet: > 3. The width: 900px applied to the Grid Col seems to create unnecessary empty space.

This is fixed, coming in next commit

aktasmehmet: > 3. On the Design Themes page, some components overlap, and a horizontal scrollbar appears when they don't fit. If we stack the components vertically instead of forcing them to stay side by side, we can avoid the horizontal scrollbar and also fix the overlapping issue.

Including this, all of those are fixed.

Your last screenshot is not fixed and it will be as soon as the body portal overlays are introduced in a couple of days.

hatesalp: Ok, I checked it. It looks fixed.

hatesalp (approved):

PR #162, opened 2026-09-15
fix(Offcanvas): Offcanvas component creation

atakanbayrak:
## Description

Offcanvas component creation is completed with test and stories file.

## Type of Change

- [x] 🆕 New Component
- [ ] ✨ Feature / Enhancement
- [ ] 🐛 Bug Fix
- [ ] 💥 Breaking Change
- [ ] 🧹 Code Refactoring

## Screenshots / Preview

Screenshot 2026-09-14 at 11 45 03 (2)
Screenshot 2026-09-14 at 11 45 31 (2)
Screenshot 2026-09-14 at 11 45 39 (2)
Screenshot 2026-09-14 at 11 45 46 (2)

## Checklist

## How to Test

[page.tsx](https://github.com/user-attachments/files/32226994/page.tsx)


src/lib/styles/tokens/design-tokens.tokens.json
```diff
@@ -1 +1 @@
...
...value":20},"paragraphIndent":{"type":"dimension","value":0},"paragraphSpacing":{"type":"dimension","value":0},"textCase":{"type":"string","value":"none"}},"xs":{"fontSize":{"type":"dimension","value":12},"textDecoration":{"type":"string","value":"none"},"fontFamily":{"type":"string","value":"Inter"},"fontWeight":{"type":"number","value":400},"fontStyle":{"type":"string","value":"normal"},"fontStretch":{"type":"string","value":"normal"},"letterSpacing":{"type":"dimension","value":0},"lineHeight":{"type":"dimension","value":18},"paragraphIndent":{"type":"dimension","value":0},"paragraphSpacing":{"type":"dimension","value":0},"textCase":{"type":"string","value":"none"}},"xxs":{"fontSize":{"type":"dimension","value":11},"textDecoration":{"type":"string","value":"none"},"fontFamily":{"type":"string","value":"Inter"},"fontWeight":{"type":"number","value":400},"fontStyle":{"type":"string","value":"normal"},"fontStretch":{"type":"string","value":"normal"},"letterSpacing":{"type":"dimension","value":0},"lineHeight":{"type":"dimension","value":16},"paragraphIndent":{"type":"dimension","value":0},"paragraphSpacing":{"type":"dimension","value":0},"textCase":{"type":"string","value":"none"}}}}}
```
atakanbayrak, line 1: Image

src/lib/components/Offcanvas/Offcanvas.test.tsx
```diff
@@ -0,0 +1,117 @@
+import Offcanvas from "@/components/Offcanvas/Offcanvas";
+import { fireEvent, render, screen, cleanup, act } from "@testing-library/react";
+import { Size3 } from "../../types";
+import { OffcanvasPosition } from "./types";
+import { userEvent } from "@testing-library/user-event";
+
+describe("Offcanvas", () => {
+  it("should render with only required props", () => {
+    expect(render(<Offcanvas open>Test content</Offcanvas>).container).toMatchSnapshot();
```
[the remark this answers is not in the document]
atakanbayrak: I implemented this fix to both Offcanvas and Modal.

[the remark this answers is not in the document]
atakanbayrak: Not necessary.

src/lib/components/Offcanvas/Offcanvas.test.tsx
```diff
@@ -0,0 +1,123 @@
+import Offcanvas from "@/components/Offcanvas/Offcanvas";
+import { fireEvent, render, screen, cleanup, act } from "@testing-library/react";
+import { Size3 } from "../../types";
+import { OffcanvasPosition } from "./types";
+import { userEvent } from "@testing-library/user-event";
+
+describe("Offcanvas", () => {
+  it("should be rendered with only required props and should have default prop values", () => {
+    render(<Offcanvas open>Test Content</Offcanvas>);
```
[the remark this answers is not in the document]
atakanbayrak: This is my design decision, I don't wanna be bothered by you again.

src/lib/components/Offcanvas/Offcanvas.tsx
```diff
@@ -0,0 +1,74 @@
...
+    ref,
+    style,
+    className,
+  } = usePropsWithThemeDefaults("Offcanvas", props);
+
+  const domReady = useDomReady();
+  const [visible, setVisible] = useState(open);
+  const [attached, setAttached] = useState(open);
+
+  const handleCloseWithAnimation = useCallback(() => {
```
hatesalp: The `AlertModal `component I’m working on had the same structure. At @aktasmehmet  request, we changed it to use `useToggle`. Once that part is clear, we may need to check this again. I’m leaving this note here for now.

src/lib/components/Offcanvas/Offcanvas.module.scss
```diff
@@ -0,0 +1,162 @@
...
+
+.headerTitle {
+  flex: 1;
+  color: var(--theme-color-text-base-title);
+  overflow-wrap: break-word;
+}
+
+.closeButton {
+  flex-shrink: 0;
+  margin-left: auto;
```
ZehranurC, line 97: You could probably remove this. `.headerTitle `having `flex: 1` already pushes the button to the end.

src/lib/components/Offcanvas/Offcanvas.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+      jest.advanceTimersByTime(300);
+    });
+    expect(handleClose).toHaveBeenCalled();
+
+    rerender(
+      <Offcanvas open closable={false} onClose={handleClose}>
+        Test content
+      </Offcanvas>,
+    );
+    const handleClose2 = jest.fn();
```
ZehranurC: handleClose2 is never wired to the Offcanvas component. You should review this test.

ZehranurC: In fact, this same issue exists in the similar test for Modal as well. there too, handleClose2 is never wired to the component. Could you check that one too?

src/lib/components/Modal/Modal.test.tsx
ZehranurC: This is exactly the place I meant, regarding handleClose2 not being wired to the component in Offcanvas.

src/lib/components/Offcanvas/Offcanvas.tsx
```diff
@@ -0,0 +1,74 @@
...
+  return (
+    attached &&
+    domReady &&
+    createPortal(
+      <div data-testid="offcanvasBackdrop" className={classNames} style={style} ref={ref}>
+        <div className={styles.offcanvas} ref={offcanvasRef}>
+          {(title || closable) && (
+            <div className={styles.header}>
+              {title && <span className={styles.headerTitle}>{title}</span>}
+              {closable && <MotifIconButton name="close" onClick={onClose} size={size} className={styles.closeButton} />}
```
ZehranurC:
```suggestion
              {closable && <MotifIconButton name="close" onClick={handleCloseWithAnimation} size={size} className={styles.closeButton} />}
```

ZehranurC: the close button's onClick calls onClose directly instead of handleCloseWithAnimation, which probably causes onClose to be triggered twice on every click (once immediately, and once via the useEffect). It might be better to use `handleCloseWithAnimation` instead of onClose here. Could you check this?

ZehranurC (changes requested):

src/lib/components/Offcanvas/Offcanvas.module.scss
```diff
@@ -0,0 +1,162 @@
...
+  ("lg", "400x")
+);
+
+$horizontal-sizes: (
+  //sizeName, height
+  ("sm", "216x"),
+  ("md", "256x"),
+  ("lg", "320x")
+);
+
```
hatesalp:
```
$sizes: (
  // sizeName, width, height
  ("sm", "320x", "216x"),
  ("md", "360x", "256x"),
  ("lg", "400x", "320x")
);
@each $size, $width, $height in $sizes {
  &.left.#{$size} .offcanvas,
  &.right.#{$size} .offcanvas {
    width: var(--base-sizing-#{$width});
  }

  &.top.#{$size} .offcanvas,
  &.bottom.#{$size} .offcanvas {
    height: var(--base-sizing-#{$height});
  }

  &.#{$size} .headerTitle {
    @include font("base", $size, "700");
  }
}
```

The two separate size maps ($vertical-sizes, $horizontal-sizes) and the standalone $header-title-sizes list can be merged into a single $sizes variable with three values per entry (name, width, height). This removes the duplicate size name definitions and consolidates all size-related rules into one loop.

src/lib/components/Offcanvas/Offcanvas.tsx
hatesalp: The close button renders at 20px but Figma specifies 18px for the sm size. It is worth checking.

src/lib/components/Offcanvas/Offcanvas.tsx
```diff
@@ -0,0 +1,74 @@
...
+    open,
+    onClose,
+    children,
+    position = "left",
+    size = "md",
+    closable = true,
+    ref,
+    style,
+    className,
+  } = usePropsWithThemeDefaults("Offcanvas", props);
```
hatesalp, line 29: A new prop can be added to render the Offcanvas without the darkened overlay backdrop.

src/lib/components/Offcanvas/Offcanvas.module.scss
```diff
@@ -0,0 +1,162 @@
...
+  flex-direction: column;
+  position: fixed;
+  background: var(--theme-color-surface-base-global-light);
+  border: solid var(--base-sizing-1x) var(--theme-color-border-light-default);
+  max-width: 100%;
+  max-height: 100%;
+  overflow: hidden;
+  @include shadow(4);
+  @include transition(transform);
+}
```
hatesalp, line 79: The position and show modifier classes are currently applied to the Root element, which forces the SCSS to use descendant selectors like .left .offcanvas or .left.show .offcanvas.  I am adding the code below as an example.


```
.offcanvas {
.....

  .left & {
      ......
  }
  .left .show &{
     .....
  }
 .left.show,
  .right.show & {
 .....
  }

  .top.show,
  .bottom.show & {
.....
}
```

hatesalp (dismissed):

src/lib/components/Offcanvas/Offcanvas.tsx
```diff
@@ -0,0 +1,74 @@
+"use client";
+
```
hatesalp, line 2: I showed this when we were sitting together. In mobile view, when the page has scroll, there is an issue with the top and bottom positions after scrolling.

src/lib/components/Offcanvas/Offcanvas.module.scss
atakanbayrak: title is optional (types.ts:14), and Offcanvas.tsx:50 renders the header when title || closable — so a closable offcanvas with no title renders .closeButton alone in .header, with no .headerTitle sibling to provide the flex: 1 that pushes it to the end.

 In that case margin-left: auto on .closeButton is what keeps the button right-aligned; without it, it would sit flush left (default flex-start). So I'd push back on removing the whole rule — it's not fully redundant, just redundant in the common "title + close button" case. I'd keep .closeButton { flex-shrink: 0; margin-left: auto; } as-is: flex-shrink: 0 also guards against the icon button shrinking if the title is long, which flex: 1 on .headerTitle alone doesn't prevent.

src/lib/components/Offcanvas/Offcanvas.tsx
atakanbayrak: We will use ViewTransition instead of handleCloseWithAnimation.

src/lib/components/Offcanvas/Offcanvas.module.scss
atakanbayrak: We have to pick one of the two: either go back to three separate size maps and get the SCSS/selector side cleaned up as suggested, or keep the single consolidated size map on the CSS side and leave the selector structure as it currently is (nested under .Root) — we can't have both at the same time, since the two rule sets live under different selectors (.offcanvas vs .Root, since .headerTitle isn't a descendant of .offcanvas).

src/lib/components/Offcanvas/Offcanvas.tsx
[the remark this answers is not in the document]
atakanbayrak: We solved this problem with fresh announced ViewTransition API at React version 19.3

src/lib/components/Offcanvas/Offcanvas.tsx
atakanbayrak: Screenshot 2026-09-17 at 10 31 04
Screenshot 2026-09-17 at 10 31 39

I hope everything seems OK rn.

src/lib/components/Offcanvas/Offcanvas.module.scss
ZehranurC: +1

src/lib/components/Offcanvas/Offcanvas.module.scss
```diff
@@ -0,0 +1,127 @@
...
+  position: fixed;
+  inset: 0;
+  width: 100%;
+  height: 100%;
+  min-height: 100%;
+  background-color: var(--base-colors-black-black-overlay50);
+  display: flex;
+  z-index: 999;
+
+  &.hideBackdrop {
```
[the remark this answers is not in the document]
atakanbayrak: Better to discuss as a team to this suggestion. @motif-ui/core-team. In my opinion, visual transparency is enough for Offcanvas.

src/lib/components/Offcanvas/Offcanvas.tsx
```diff
@@ -0,0 +1,68 @@
...
+const Offcanvas = (props: PropsWithRef<OffcanvasProps, HTMLDivElement>) => {
+  const {
+    title,
+    open,
+    onClose,
+    children,
+    position = "left",
+    size = "md",
+    closable = true,
+    hideBackdrop = false,
```
hatesalp: Are the default values for `hideBackdrop `and `closable `necessary here?

